PR #6, opened 2021-12-08
Lighthouse import

gbayasgalan:


controls/roles/validator-import/molecule/default/prepare.yml
```diff
@@ -4,7 +4,9 @@
   vars_files:
     - ../../../../defaults/stereum_defaults.yaml
   vars:
-    eth1_node: https://goerli.infura.io/v3/20fd3ed2418742e9b727857e32b40f9c  
+    eth1_node: https://goerli.infura.io/v3/20fd3ed2418742e9b727857e32b40f9c
```
stefa2k, line 7: we shouldn't put this here, it's actually a secret... do we need this in order for the test to work?

stefa2k (changes requested):

controls/roles/validator-import/molecule/default/prepare.yml
gbayasgalan: okey, got it

PR #34, opened 2022-01-14
Lighthouse-SSV

gbayasgalan:


controls/roles/manage-service/tasks/main.yml
```diff
@@ -37,6 +37,7 @@
       owner: "{{ stereum_service_configuration.user }}"
       group: "2000"
       mode: 0700
+    ignore_errors: yes
```
stefa2k, line 40: I think a better, safer, approach would be to:
- check if the path is a file or a directory
- set owner, group, mode for file (when file)
- set owner, group, mode for directory (when directory)

controls/roles/manage-service/tasks/ssv-lighthouse/molecule/default/prepare.yml
```diff
@@ -0,0 +1,112 @@
...
+          manage_service:
+            save: true
+            state: started
+            configuration:
+              id: "{{ ssv_service }}"
+              image: bloxstaking/ssv-node:latest-ubuntu
+              ports:
+                - 0.0.0.0:12000:12000/udp
+                - 0.0.0.0:13000:13000/tcp
+              command: make BUILD_PATH=/go/bin/ssvnode start-node && docker logs ssv_node
```
stefa2k, line 105: The goal is to bring the creation of the config file into the command, so we don't need an extra task for the config file creation.

If this is not possible in any way then we need to discuss how to handle config files in the `manage-service` role properly.

stefa2k (changes requested):
Directory structure is incorrect, current position:
https://github.com/gbayasgalan/ethereum-node/tree/ssv/controls/roles/manage-service/tasks

It should be put here, like the molecule scenarios `start-multiple`, `start-stop`, ...
https://github.com/gbayasgalan/ethereum-node/tree/ssv/controls/roles/manage-service/molecule

controls/roles/manage-service/tasks/main.yml
gbayasgalan: yes, that would be best option, I'll have look

gbayasgalan: solved using env & command, no need to mount "config file to file"

controls/roles/manage-service/tasks/ssv-lighthouse/molecule/default/prepare.yml
gbayasgalan: solved using env & echo

controls/roles/manage-service/molecule/ssv-lighthouse/prepare.yml
```diff
@@ -0,0 +1,113 @@
...
+                - 127.0.0.1:5052:5052/tcp
+              command: []
+              entrypoint: ["/opt/app/start/beacon.sh"]
+              user: "2000"
+              volumes:
+                - "/opt/app/services/{{ beacon_service }}/beacon:/opt/app/beacon"
+                - "/opt/app/services/{{ beacon_service }}/slasher:/opt/app/slasher"
+
+    # ssv service
+    - block:
```
stefa2k, line 76: I think this block should be in `converge.yaml`, because the test is about SSV.

controls/roles/manage-service/molecule/ssv-lighthouse/verify.yml
```diff
@@ -0,0 +1,17 @@
...
+  tasks:
+  - stat: path=/etc/stereum/services
+    register: stereum_services_dir
+  - debug:
+      msg: "{{ stereum_services_dir }}"
+  - shell: docker ps
+    register: stereum_docker_ps
+  - debug:
+      msg: "{{ stereum_docker_ps }}"
+  - assert:
```
stefa2k, line 14: Can you verify also the logs?

controls/roles/manage-service/tasks/main.yml
```diff
@@ -36,7 +36,7 @@
       state: directory
       owner: "{{ stereum_service_configuration.user }}"
       group: "2000"
-      mode: 0700
+      mode: 0700
```
stefa2k, line 39: please remove trailing spaces, lint doesn't like them

stefa2k (changes requested):

controls/roles/manage-service/molecule/ssv-lighthouse/verify.yml
gbayasgalan: yes sure, I will

controls/roles/manage-service/tasks/main.yml
gbayasgalan: it is removed

controls/roles/manage-service/molecule/ssv-lighthouse/prepare.yml
gbayasgalan: done!

controls/roles/manage-service/molecule/ssv-lighthouse/verify.yml
gbayasgalan: done!

controls/roles/manage-service/tasks/main.yml
```diff
@@ -34,9 +34,10 @@
     file:
       path: "{{ item.split(':') | first }}"
       state: directory
-      owner: "{{ stereum_service_configuration.user }}"
+      owner: "2000"
```
stefa2k, line 37: Is there a reason for forcing "2000" instead of using `{{ stereum_service_configuration.user }}`? Grafana for example needs to run as root, otherwise it doesn't work properly.

stefa2k (changes requested):

controls/roles/manage-service/tasks/main.yml
gbayasgalan: you right, there is a no need to change this line, since the service users are "2000". Now users given as parameter

stefa2k (approved):

PR #48, opened 2022-01-31
New tests

maxbehzadi82:


launcher/tests/unit/layers/InstallOption.spec.js
```diff
@@ -0,0 +1,9 @@
+import { mount } from "@vue/test-utils";
+import InstallOption from "../../../src/components/layers/InstallOption";
+describe("InstallOption.vue", () => {
+  test("renders a router-link tag with to temp.url", () => {
+    const wrapper = mount(InstallOption);
+
+    expect(wrapper.find("router-link").exists()).toBe(true);
+  });
+});
```
NeoPlays, line 9: You might want to check if the list in `Installoption.vue` gets rendered because it contains images which are extern resources and can cause errors (could not find file ... etc). I hope you know what i mean 😅

NeoPlays (commented):
LGTM 👍🏼 just let me know what you think about my comment on `InstallOptions.spec.js`

launcher/tests/unit/layers/InstallOption.spec.js
maxbehzadi82: Of course we can.As stefan said,at the moment we only need totally basic tests.

NeoPlays (approved):

launcher/tests/unit/layers/InstallOption.spec.js
maxbehzadi82: Well done bud.

PR #55, opened 2022-02-06
FIX:molecule static geth & docker log verify

stefa2k:
This PR addresses multiple issues:
- No need for infura.io eth1 addresses on github action hetzner vms, there are 2 Geth instances running on a private network `10.10.0.0/24` the test VMs can join. `10.10.0.2` for websocket and `10.10.0.3` for RESTful endpoints.
- Checking output of logs of running services in `stderr` instead of `stdout`. `docker logs` always writes to `stderr`, e. g. `docker ps` writes like everything else to `stdout´.


gbayasgalan (approved):
lgtm

PR #64, opened 2022-02-08
ADD: Hetzner Server Creation Class, Integration Test

NeoPlays:


.github/workflows/test-jest.yml
```diff
@@ -20,3 +20,5 @@ jobs:
         working-directory: ./launcher
       - run: npm run test
         working-directory: ./launcher
+        env:
+          HCLOUD_TOKEN: ${{ secrets.HCLOUD_TOKEN }}
```
stefa2k, line 24: can we get a separate github action for JEST integration tests? The new JEST integration test job shouldn't run on PRs, just on merges on branch `main` like the molecule tests. This is because the `secrets.HCLOUD_TOKEN` could be exposed by an attacker just doing something like `log.debug(HCLOUD_TOKEN)`.

launcher/src/backend/HetznerServer.js
```diff
@@ -0,0 +1,159 @@
...
+const https = require('https');
+
+/**
+ * Used to wait for the server to fully initialize
+ */
+function Sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+export class HetznerServer {
```
stefa2k, line 11: Can we put this class on an integration test file so it's only loaded for integration tests?

launcher/src/backend/NodeConnection.int.js
```diff
@@ -0,0 +1,69 @@
...
+jest.setTimeout(500000);
+
+function Sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+test('prepareStereumNode on ubuntu', async () => {
+
+    const serverSettings = {
+        name: 'my-server',
```
stefa2k, line 16: Format for server name is with molecule: `<ansible-role>--<scenario-name>--<os-name>`

Can you use something like this please? It makes it easier to identify stuck machines on Hetzner later.

stefa2k (changes requested):

launcher/jest.config.js
```diff
@@ -6,9 +6,9 @@ module.exports = {
   ],
   roots: ["<rootDir>/src/", "<rootDir>/tests/"],
   // testEnvironment: "jest-environment-jsdom",
-  // testMatch: [
-  //   "**/?(*.)+(spec|test).[tj]s?(x)"
-  // ],
+  testMatch: [
+     "**/?(*.)+(spec|test|int).[tj]s?(x)"
```
NeoPlays, line 10: added `.int` as suffix to run tests separately because I did not found a way to seperate `int.test.js` from `test.js`

stefa2k (approved):

PR #69, opened 2022-02-11
NodeSidebar & children comps + new icons

maxbehzadi82:


launcher/src/components/UI/NodeBg.vue
```diff
@@ -1,5 +1,4 @@
 <template>
-
   <div id="parent">
```
mabasian, line 2: just using the parent name will make change to all elements in all componenmts, please make a more clear name maybe sidebar-parent

mabasian (approved):
just more clear name to parent container

PR #74, opened 2022-02-18
ADD: Wiring to BE for FormSetup, fixed FormSetup

NeoPlays:
it's not the prettiest but it works. I also had an error saying `preload.js was not loaded` or something like that. I fixed it by manually copying `preload.js` into `launcher/dist_electron`.

### DIFF
--- a/launcher/jest.config.js
+++ b/launcher/jest.config.js
@@ -9,6 +9,9 @@ module.exports = {
   testMatch: [
      "**/?(*.)+(spec|test|int).[tj]s?(x)"
    ],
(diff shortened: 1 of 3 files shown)


NeoPlays:
```
FAIL tests/unit/pages/BaseHome.spec.js
  ● Test suite failed to run

    Cannot find module '@/store/ControlService' from 'src/components/layers/FormSetup.vue'
```
How is this a thing? When compiling and running the application everything works fine. @maxbehzadi82 do you have an idea?

PR #78, opened 2022-02-20
FIX:local settings overlapping stereum1.0

stefa2k:
Prevent Stereum 2.x from reading/writing Stereum 1.x local settings.


NeoPlays (approved):
legit 😅 didn't thought of that

PR #82, opened 2022-02-23
FIX: Server getting stuck on integration test

NeoPlays:
Test failed on too many tries to connect. The server got stuck and prevented re-runs from being successful


stefa2k (approved):
great catch!

PR #89, opened 2022-02-27
ADD: Nimbus monitoring test (prometheus, prom-node-exporter and grafana)

gbayasgalan:


controls/roles/manage-service/molecule/monitor-nimbus/converge.yml
```diff
@@ -0,0 +1,114 @@
...
+- name: Converge
+  hosts: all
+  vars_files:
+  - ../../../../defaults/stereum_defaults.yaml
+  vars:
+    prometheus_service: 4a87cc9c-9744-11ec-a2ef-73b1b3e1ee0e
+    prometheus_node_exporter_service: 77784ab0-9744-11ec-9726-ab848289b406
+    grafana_service: 52134fcc-9744-11ec-b5ca-572497326b43
+    beacon_service: 5ac8408c-9744-11ec-8b66-0f11f6579737
+    grafana_provisioning: provisioning-n
```
stefa2k, line 11: I'd suggest using full names here to prevent confusion. `provisioning-n` could mean `nimbus` or `nethermind` or `network` or whatever. How about `nimbus` without the `provisioning-` prefix which we don't need anyway because of the variable name?

stefa2k (changes requested):

controls/roles/manage-service/molecule/monitor-nimbus/converge.yml
gbayasgalan: changed: "provisioning-lh -> lighthouse" and "provisioning-n -> nimbus"

stefa2k (approved):
lgtm

PR #132, opened 2022-03-24
FIX: genericPlaybook.yaml throwing error on undefined args

NeoPlays:
Role `setup` doesn't require any `stereum_args` and uses `stereum_static` instead, but `combine()` throws an error on undefined values.

### DIFF
--- a/controls/genericPlaybook.yaml
+++ b/controls/genericPlaybook.yaml
@@ -8,6 +8,9 @@
     - set_fact:
         stereum: "{{ stereum_static | combine(stereum_args, recursive=True) }}"
       when: stereum_args is defined
+    - set_fact:
+        stereum: "{{ stereum_static}}"
+      when: stereum_args is undefined
     - name: "Include role"
       include_role:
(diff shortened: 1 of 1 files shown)


controls/genericPlaybook.yaml
```diff
@@ -7,6 +7,7 @@
   tasks:
     - set_fact:
         stereum: "{{ stereum_static | combine(stereum_args, recursive=True) }}"
+      when: stereum_args is defined
```
stefa2k, line 10: shouldn't there be a
```
- set_fact:
    stereum: "{{ stereum_static}}"
  when: stereum_args is undefined
```
otherwise var `stereum` would be empty if `stereum_args` isn't set

NeoPlays: `is defined` is fine because it should only be combined if there is something to combine, but i'll change it so `stereum` is set to `stereum_static` if `stereum_args` is undefined to prevent any errors with `stereum` being undefined

stefa2k (approved):

PR #138, opened 2022-03-31
FIX: prysm ssv test verify

stefa2k:
### Possible issue
Blox SSV connects to beacon node successfully and runs for a while before we verify the log of the SSV service. Because we only verify the last 250 lines of the output this could push the strings in the output we are looking for too much up (out of scope of the `docker logs --tail=250`).

### Solution
- check Blox SSV service to be connected with healthy beacon node
- check Blox SSV service to have the operator public keys loaded


NeoPlays (approved):
lgtm

PR #195, opened 2022-04-15
FIX: issue with bundler renaming class constructor

NeoPlays:


launcher/src/backend/ethereum-services/GethService.js
```diff
@@ -9,6 +9,7 @@ export class GethService extends NodeService {
 
         const service = new GethService();
         service.init(null,
+            "GethService",
```
stefa2k, line 12: shouldn't there be another `null` as argument for the id?

stefa2k (changes requested):

launcher/src/backend/ethereum-services/GethService.js
stefa2k: or switch `null` with `"GethService"` one line up?

maxbehzadi82 (dismissed):
lg

launcher/src/backend/ethereum-services/GethService.js
NeoPlays: ohhh man and i even tripple checked 😅

stefa2k (approved):
thx

stefa2k (approved):

PR #262, opened 2022-05-11
fix: checking ssv log output

gbayasgalan:
with `v0.1.12`, There is a no Information about ssv-keys in Log output


stefa2k (changes requested):
Is there no information to verify the ssv-node actually started up? Maybe some log line that is pretty common? Maybe check at least for

```
2022-05-10T21:54:00.000273Z     DEBUG   duties/fetcher.go:127   no indices, duties won't be fetched     {"app": "SSV-Node:v0.1.12", "component": "operator/dutyFetcher"}
```

something like `no indices, duties won't be fetched` should be enough to check for

stefa2k (approved):

PR #270, opened 2022-05-12
ADD: simple stereum update process

stefa2k:
Simple update produre for Stereum controls on the node itself. Also 2 tests for update to latest and update to a specific version (via git commit hash) included.

### DIFF
--- a/controls/roles/update-stereum/tasks/main.yml
+++ b/controls/roles/update-stereum/tasks/main.yml
@@ -1,0 +1,32 @@
+---
+- name: Download update metadata
+  uri:
+    url: https://stereum.net/downloads/updates.json
+    method: GET
+    return_content: true
+    status_code: 200
+    body_format: json
+  register: update_data_download
+
+- name: Set update data
+  set_fact:
+    update_data: "{{ update_data_test | default(update_data_download) }}"
+
+- name: Fetch repository information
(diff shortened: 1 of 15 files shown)


controls/roles/update-stereum/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+  register: update_data_download
+
+- name: Set update data
+  set_fact:
+    update_data: "{{ update_data_test | default(update_data_download) }}"
+
+- name: Fetch repository information
+  shell: git fetch --all
+  args:
+    chdir: "{{ stereum.settings.controls_install_path }}"
```
NeoPlays, line 18: `stereum.settings.controls_install_path` is by default `/opt/stereum`

`/ansible` is added in Javascript Backend when the repository is pulled.
Therefore does `stereum.settings.controls_install_path` never include `/ansible`

So i would suggest an extra variable or manipulating the path

NeoPlays (changes requested):

controls/roles/update-stereum/tasks/main.yml
stefa2k: nice catch! done

NeoPlays (approved):

PR #271, opened 2022-05-12
FIX: use default teku data storage mode

### LINKED ISSUE #170: Teku: Data storage mode
With Stereum 1.x we used `archive` data storage mode which is using a lot of disk space: https://github.com/stereum-dev/ethereum2-docker-compose/blob/master/config/teku/beacon.yaml#L25

Try to use `prune` instead and make tests to verify it's working properly with the rest of the services: https://docs.teku.consensys.net/en/latest/Reference/CLI/CLI-Syntax/#data-storage-mode

stefa2k:
resolves #170


gbayasgalan (approved):
lgtm

PR #272, opened 2022-05-12
ADD: config version for service yamls

### LINKED ISSUE #137: Service configuration file versioning
Configuration files of services located in `/etc/stereum/services` should include a version property to handle older versions of configuration versions with newer stereum launcher versions.

stefa2k:
This is necessary to update config files when breaking changes happen for Stereum versions as well as service versions.

fixes #137


NeoPlays (approved):
👍🏼

PR #284, opened 2022-05-13
FIX: ssv-prysm eth1 link

gbayasgalan:
error: `could not connect to the eth1 client` https://github.com/stereum-dev/ethereum-node/runs/6425742182?check_suite_focus=true#step:8:1121


maxbehzadi82 (approved):
looks good

PR #289, opened 2022-05-16
ADD: service-id to service-paths

NeoPlays:


launcher/src/backend/ethereum-services/GethService.js
```diff
@@ -3,24 +3,30 @@ import { ServiceVolume } from './ServiceVolume.js'
 
 export class GethService extends NodeService {
   static buildByUserInput (network, ports, workingDir) {
+    const service = new GethService()
+    service.setId()
+    workingDir = workingDir + '-' + service.id
```
stefa2k, line 8: Should be checked for trailing `/`, e. g. `workingDir = "/opt/foo/bar/"` would produce `workingDir = "/opt/foo/bar/-somekindofid"`.

How about a method to do this so it's done the same way for all services?

stefa2k (changes requested):

stefa2k (approved):

PR #292, opened 2022-05-18
ADD: Prysm tests, fix smaller issues

NeoPlays:


controls/roles/validator-import-api/tasks/main.yml
```diff
@@ -2,7 +2,7 @@
 - block:
   - name: Get validator service's auth-token
     slurp:
-      src: "/opt/app/services/{{ validator_service }}/data/wallets/auth-token"
+      src: "{{ auth_token_path }}"
```
NeoPlays, line 5: passing full path instead of validator id because of variable service paths.
i think there are no resources that i could extract that information from

stefa2k: Either Ansible pulls the information from the volumes or JS does it. You can look up the volumes, look for where the volume is set to and then build the correct path with the information.

NeoPlays: I didn't want to make huge changes but i'd prefer doing it in ansible

stefa2k: I'll have a look at it!

stefa2k: see https://github.com/NeoPlays/ethereum-node/pull/1

NeoPlays: Ahhh perfect, thanks a lot 🙌🏼

PR #298, opened 2022-05-19
FIX: playbook throwing warning

NeoPlays:


controls/roles/validator-import-api/tasks/main.yml
```diff
@@ -8,10 +8,10 @@
     set_fact:
       validator_service_config: "{{ validator_service_config_raw.content | b64decode | from_yaml }}"
   - name: Build api token path
```
stefa2k, line 10: shouldn't this be like this?
```
  - name: Extract service volumes
    set_fact:
      wallet_image: "{{ validator_service_config.volumes | select('search', ':/opt/app/data/wallets') | first}}"
  - name: Build token path
    set_fact:
      auth_token_path: "{{ wallet_image.split(':') | first }}/auth-token"
```

stefa2k (changes requested):

controls/roles/validator-import-api/tasks/main.yml
NeoPlays: In my test playbook it woked, but i'll change it up. looks better 👍🏼

stefa2k (approved):

PR #305, opened 2022-05-20
FIX: PrysmBeaconService.int.js

NeoPlays:


launcher/src/backend/ethereum-services/PrysmBeaconService.int.js
```diff
@@ -151,7 +153,7 @@ test('prysm validator import', async () => {
     //expect(BCstatus.stderr).not.toMatch(/Could not connect to execution endpoint/)
 
     //check prysm VC logs
-    expect(VCstatus.stderr).toMatch(/Validating for public key/)
     expect(VCstatus.stderr).toMatch(/Beacon chain started/)
+    expect(runningValidator).toMatch(/Showing 3 validator accounts/)
```
gbayasgalan, line 157: `3` is not fixed number of validators

NeoPlays: that's true but i'm only importing 3 so if it's not 3 something did not work

launcher/src/backend/ethereum-services/PrysmBeaconService.int.js
```diff
@@ -118,10 +118,12 @@ test('prysm validator import', async () => {
     const ufw = await nodeConnection.sshService.exec('ufw status')
     const validatorAccounts = await nodeConnection.sshService.exec(`cat ${wallet_path}/direct/accounts/all-accounts.keystore.json`)
     const auth_token = await nodeConnection.sshService.exec(`cat ${wallet_path}/auth-token`)
-    const BCstatus = await nodeConnection.sshService.exec(`docker logs --tail=100 stereum-${prysmBC.id}`)
-    const VCstatus = await nodeConnection.sshService.exec(`docker logs --tail=70 stereum-${prysmVC.id}`)
+    const BCstatus = await nodeConnection.sshService.exec(`docker logs stereum-${prysmBC.id}`)
```
gbayasgalan, line 121: so far I know, It'll not work, at least prysm-VC. `--tail=70` for prysm-vc is optimal number of lines that we can check. But it depends on which device speed, for example on github test it work not everytime, but on my laptop it works well

launcher/src/backend/ethereum-services/PrysmBeaconService.int.js
gbayasgalan: okey I see. then should be fine for this case

launcher/src/backend/ethereum-services/PrysmBeaconService.int.js
NeoPlays: that are the issues I'm having too, running everything from JavaScript Backend also makes huge differences
So my idea was to run it on GH and look at all logs and then adapt.

gbayasgalan (approved):
lgtm

Issue #213, opened 2022-04-20
Replace stereum/lighthouse docker images in molecule tests


stefa2k:

gbayasgalan:
https://github.com/stereum-dev/ethereum-node/pull/328

Issue #210, opened 2022-04-20
Use original consensus client images


stefa2k:
Right now we make use of stereum built docker images, like `stereum/nimbus` and `stereum/lighthouse`. Please replace these with the original consensus client images.

- [x] #211
- [x] #212
- [ ] #213
- [ ] #214

gbayasgalan:
https://github.com/stereum-dev/ethereum-node/pull/328

PR #306, opened 2022-05-20
FIX: replace stereum/* docker images in molecule tests & EDIT: CLI import is deactivated

gbayasgalan:
https://github.com/stereum-dev/ethereum-node/issues/214
https://github.com/stereum-dev/ethereum-node/issues/213
https://github.com/stereum-dev/ethereum-node/issues/210

### DIFF
--- a/controls/roles/validator-import-lighthouse/tasks/main.yml
+++ b/controls/roles/validator-import-lighthouse/tasks/main.yml
@@ -23,16 +23,17 @@
       }
 
       spawn sh -c { sudo docker run --rm -it \
-                     -e DEBUG_LEVEL=debug \
-                     -e NETWORK=prater \
-                     -e DATADIR="/opt/app/validator" \
-                     -e DIRECTORY="/opt/app/launchpad" \
-                     -v "/opt/app/services/{{ validator_service_name }}/validator":"/opt/app/validator" \
-                     -v "/opt/app/services/{{ validator_service_name }}/launchpad":"/opt/app/launchpad" \
+                     -v $HOME/.lighthouse:/root/.lighthouse \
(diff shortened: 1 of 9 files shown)


controls/roles/validator-import-lighthouse/tasks/main.yml
```diff
@@ -23,16 +23,17 @@
...
 
       spawn sh -c { sudo docker run --rm -it \
-                     -e DEBUG_LEVEL=debug \
-                     -e NETWORK=prater \
-                     -e DATADIR="/opt/app/validator" \
-                     -e DIRECTORY="/opt/app/launchpad" \
-                     -v "/opt/app/services/{{ validator_service_name }}/validator":"/opt/app/validator" \
-                     -v "/opt/app/services/{{ validator_service_name }}/launchpad":"/opt/app/launchpad" \
+                     -v $HOME/.lighthouse:/root/.lighthouse \
+                     -v "/opt/app/services/2e83d610-48b8-11ec-802a-330dee8e6b2e/validator":"/opt/app/validator" \
```
stefa2k, line 27: This is a fixed path, it will break as soon as this actually get's used. However, I don't think we need those CLI imports anymore.

What do you think about deactivating all tests regarding CLI import? It looks like import via API works anyway.

stefa2k (commented):
How about deactivating CLI validator-import tests? See comment. Would safe us some time right?

controls/roles/validator-import-lighthouse/tasks/main.yml
gbayasgalan: CLI import is deactivated

stefa2k (approved):

PR #318, opened 2022-05-24
FIX: load services when logging into existing node

NeoPlays:


stefa2k (commented):
lgtm, someone else should look into this too

maxbehzadi82 (approved):
looks good

PR #325, opened 2022-05-25
FIX: api nimbus verify

gbayasgalan:
https://github.com/stereum-dev/ethereum-node/runs/6590715793?check_suite_focus=true#step:8:499


controls/roles/validator-import-api/molecule/teku/create.yml
```diff
@@ -35,7 +35,7 @@
           - "{{ ssh_key_name }}"
         volumes: "{{ item.volumes | default(omit) }}"
         image: "{{ item.image }}"
-        location: "{{ item.location | default(omit) }}"
+        location: "hel1"
```
stefa2k, line 38: I think this needs to be changed also in all other `validator-import-api` molecule tests, right?

NeoPlays (approved):
👍🏼

PR #327, opened 2022-05-25
ADD: new flags & animation

maxbehzadi82:
New flags and styles were added to the select language page.
server connecting is animated on formSetum.


NeoPlays (approved):
lgtm

PR #365, opened 2022-06-03
ADD: check for necessary password change

### LINKED ISSUE #193: OS Check breaks when user needs to change password
In case the user specified in the login screen is forced to change the password at the ssh-login, the OS check fails and doesn't display anythin (empty string).

There should be a warning/message that login was successful but user needs to change password first.

NeoPlays:
resolves #193


maxbehzadi82 (approved):
ok

PR #381, opened 2022-06-09
ADD: patches for updates

stefa2k:
This PR introduces a way to implement patches running custom ansible roles & tasks on update via `controls/finishUpdate.yaml`. Similar process is used in Stereum version 1.x


gbayasgalan (approved):
lgtm

PR #389, opened 2022-06-12
ADD: delete-validator(s) with API for Lighthouse, Nimbus, Prysm and Teku. Execution layer - Geth

gbayasgalan:
delete-validator account(s) using API:
- Lighthouse
- Nimbus
- Prysm
- Teku

Execution layer:
- Geth


stefa2k (changes requested):
URLs for API should be generated dynamically, see #390 also.

PR #393, opened 2022-06-13
REFACTOR: Migrating from VueX to Pinia

maxbehzadi82:
ClickInstall state changed,
TheNode state changed
NodeManage state changed
UI styles fixed


NeoPlays (approved):
We migrated to pinia and everything should work just like before. Improvements will be implemented later.

PR #399, opened 2022-06-17
ADD: changing client on pluginName now is possible

### LINKED ISSUE #371: Make client choosable on one-click-installer "Staking"
Right now the setup chooses the client for the user and the user has no way to change the choice. The user should be able to change the selected client.

maxbehzadi82:
fixed #371
at the moment execution is changeable.
Validator and consensus are changeable, not automatically.


mabasian (approved):
lg

PR #415, opened 2022-06-21
FIX: molecule tests now use versions of defaults

stefa2k:
docker image versions are now managed centralized via `stereum.defaults.versions`


gbayasgalan (approved):
lgtm

Issue #370, opened 2022-06-07
No automatic import of Grafana dashboards


stefa2k:
Grafana doesn't import the necessary dashboards of the selected setup/usecase.

stefa2k:
Please provide the prometheus service with the network alias `prometheus`, see https://docs.ansible.com/ansible/latest/collections/community/docker/docker_container_module.html#parameter-networks/aliases

PR #430, opened 2022-06-22
FIX: replacing Prometheus-IDs in datasource.yml

gbayasgalan:
fix: https://github.com/stereum-dev/ethereum-node/issues/370


NeoPlays (commented):
I thought you would implement this in `write-grafana-configuration.yml` 😅
otherwise the JS Backend would need to do this
@stefa2k what do you think?

stefa2k:
> I thought you would implement this in `write-grafana-configuration.yml` 😅 otherwise the JS Backend would need to do this @stefa2k what do you think?

that looks to me only for tests, so we'd still need it in the JS backend right? Maybe we can talk about it tomorrow?

stefa2k:
As discussed today: Please use a network alias for prometheus docker container.

PR #432, opened 2022-06-22
FIX: molecule tests

stefa2k:


controls/roles/manage-service/molecule/ssv-prysm/prepare.yml
```diff
@@ -57,17 +57,6 @@
         mode: 0644
       become: yes
 
-    - block:
```
gbayasgalan, line 60: without fastsync? I thought beacon chain should be fully synced

stefa2k: I don't think this is necessary because we can check also without a fully synced beacon if the bloxssv connection works

gbayasgalan (approved):
lgtm 👍

PR #437, opened 2022-06-23
FIX: Integration Tests

NeoPlays:


launcher/src/backend/ethereum-services/LighthouseBeaconService.int.js
```diff
@@ -84,11 +84,14 @@ test('lighthouse validator import', async () => {
...
-            ]
+        stereum_role: 'validator-import-api', validator_service: lhVC.id , validator_keys:[{ 
+                name: 'batch0', 
+                passwords: ['MyTestPassword','MyTestPassword','MyTestPassword'], 
+                content: [
+                    '{"crypto": {"kdf": {"function": "scrypt", "params": {"dklen": 32, "n": 262144, "r": 8, "p": 1, "salt": "de4b32f49572c01146afb11a82c326fdc03be6cf447983daf9eb7ec0f868a116"}, "message": ""}, "checksum": {"function": "sha256", "params": {}, "message": "1ccb24f0f1469ab56cc0147dae242aab59ff360177c8ec4f710966913da839b6"}, "cipher": {"function": "aes-128-ctr", "params": {"iv": "a24857026939492f49444679544cb6bb"}, "message": "b5d944adfb65e33873c5c1b809c3c15b558821f2829cd7e9da1df65d0b78fdb6"}}, "description": "", "pubkey": "acaa51756fb445b406c9e599f3f4bec991f7799c002619566ab1fa5b70445c62f1ac6561154ca5e49d7542dbe690b96b", "path": "m/12381/3600/0/0/0", "uuid": "1ea9ed13-e3bb-4555-99d9-c5e83ab9eb67", "version": 4}',
```
stefa2k, line 91: I'm not sure how this works in JS, could be that you need to use double quotes `"` all the time and escape the double quotes of the actual key like so `\"`

stefa2k (approved):
let's try it like this and see if it works, otherwise see comment about quoting the key

PR #462, opened 2022-06-29
UPDATE: Blox ssv modal

### LINKED ISSUE #452: Blox SSV key copy doesn't work
The copy button of both keys copies the same key. It should copy the key of the specific text field.

image

MaxTheGeeek:
fixed #452
Blox ssv copy function fixed
creating secret key added
pubkey & secret key changed display


NeoPlays (approved):
lgtm 🚀

PR #463, opened 2022-06-29
ADD: Grafana provisioning in ansible

### LINKED ISSUE #450: Starting plugin doesn't work
After successfully stopping Grafana it doesn't work to start it again. The on-off button is still red too.

image

stefa2k:
resolves #450 

Please note:
- Folders in `controls/roles/manage-service/files/grafana` are named after the service property of the service's config yaml
- Provisioning by service, this means we can install dashboards with high granularity. E. g. install validator dashboards only if validator service is installed too


NeoPlays (approved):
lgtm 🚀

PR #477, opened 2022-07-01
ADD: closing dialog

### LINKED ISSUE #456: Setup breaks on closing application when installing
Closing Stereum after having started the installation on the server (and before finishing), leads to a problem when trying to login again.

NeoPlays:
grafik

When the stereum launcher is executing some critical tasks the user must confirm to exit (if the user ignores it well what can i do 😅)
fixes #456


stefa2k (approved):
makes sense 🙃

PR #493, opened 2022-07-04
FIX: set default values when necessary on setup

stefa2k:
fixing issue introduced with #492 

The default values (like installation path) are now set only when not already filled. `stereum` variable get's set by `controls/genericPlaybook.yaml` which is used by Stereum Launcher.


gbayasgalan (approved):
👍

PR #494, opened 2022-07-05
FIX: remove duplication of key "until"

stefa2k:
this should resolve the molecule lint issues for the `manage-service` role


NeoPlays (approved):
👍🏼

PR #495, opened 2022-07-05
UPDATE: OneClickInstaller

NeoPlays:
also 

- removed service data from old store
- hopefully fixed grafana for the last time
- when selecting for removal (NodeManage) more than 1 were selectable but only the last got deleted (fixed)
- non playbook and failing tasks kept the task in taskmanager running (fixed)


maxbehzadi82 (approved):
good job

PR #515, opened 2022-07-07
FIX: supporting sudo user

### LINKED ISSUE #510: Support super user setups
Right now working with a super user breaks the setup. Users should be able to use a super user instead of root to install & manage their node.

stefa2k:
fixes #510 

read up on the mechanics: https://unix.stackexchange.com/questions/175415/forward-function-and-variables-into-sudo-su-user-eof

### DIFF
--- a/launcher/src/backend/SSHService.js
+++ b/launcher/src/backend/SSHService.js
@@ -84,10 +84,6 @@ export class SSHService {
     return this.execCommand(ensureSudoCommand, logline)
   }
 
-  async execCommand (command) {
-    return this.exec(command, command)
-  }
-
   async execCommand (command, logline) {
     log.info('exec', logline)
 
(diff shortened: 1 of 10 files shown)


launcher/src/backend/SSHService.js
```diff
@@ -77,6 +79,16 @@ export class SSHService {
   }
 
   async exec (command, logline) {
+    const ensureSudoCommand = "sudo -u 'root' -i <<'=====EOF'\n" + command + "\n=====EOF"
+
+    return this.execCommand(ensureSudoCommand, logline)
+  }
+
+  async execCommand (command) {
+    return this.exec(command, command)
```
NeoPlays, line 88: shouldn't this be `return this.execCommand(command, command)`

NeoPlays (changes requested):

PR #517, opened 2022-07-08
Fix molecule prysm verify

stefa2k:
- removing test for setting fee recipient for github actions only: doesn't work because of testnet implementation not ready yet.
- checking prysm not in logs but with api call, this way log can be full of unsynced-messages/-warnings and we still know if it's working or not


gbayasgalan (approved):
lgtm 👍

PR #548, opened 2022-07-15
FIX: many issues

NeoPlays:
Fixes:

- Services with no specific validator service didn't load correctly
- no more loading of services in `FormSetup.vue` and `VerifyInstallation.vue`
- bugs while deleting config
- and other smaller problems


maxbehzadi82 (approved):
gj

PR #553, opened 2022-07-18
FIX: Small fixes bundle

stefa2k:
fixes:
- slow install of packages of ufw
- mass test creation of molecule tests -> now in tree structure with dependencies


NeoPlays (approved):
lgtm 👍🏼

PR #556, opened 2022-07-19
FIX: resolve molecule issue with apt (stuck) & remove outdated ansible roles

stefa2k:
There's an issue with interactive apt in molecule `prepare.yaml`, using ansible module `apt` instead of executing the command `apt` via `raw` solves the issue.


NeoPlays (approved):
lgtm 🚀

Issue #269, opened 2022-05-12
Update notifier Stereum Launcher


stefa2k:
Stereum launcher should check available updates of the app on startup or regularly and notify the user when updates are available to download.

@daGscheid please design how this should look like.

daGscheid:
There are two different cases of updates we have to handle:

1) Stereum Updates 
2) Plugin Updates 

To outline my thinking: For clarity there should be a space that features both to conveniently access. 
As of the topic of the issue this explanation is concentrating on the first category, updates of Stereum itself. 
The elements inside this category can than be further classified.

BEFORE LOGIN
AFTER LOGIN

In this 1# response I would like to concentrate on the update process AFTER LOGIN - given that decision made in #268 will highly influence decision made on this issue. I will summarize them again later with less blablabla in a later comment.

<br>
<br>
<br>

###BEFORE LOGIN
### NOTIFICIATION
---

Currently these are all planned spaces **`to notify`** the user of Stereum updates: 

ON ALL PAGES
--- 
![Update available Launcher](https://user-images.githubusercontent.com/82385103/169283209-ec20fd6e-b124-4011-9a42-0f9cea808571.png)

Replace left top icon with update notification. 

<br>
<br>
<br>

### INITIATION
---

Pressing the download buttons generally leads to a confirmation dialog where the version to which Stereum is updated to is presented to the user. Clicking accept makes a simple download progress bar / circle / numerals pop up. After this the Stereum would probably require a restart (@stefa2k)?
Currently these are all planned spaces where the user is able `to initiate ` the `update process of Stereum`:  


PRESSING THE LOGO NOTIFICATIONs
--- 

Pressing the notification on the logo in the top left, that is used to notify you, leads to a page with this element
![TAB MENU](https://user-images.githubusercontent.com/82385103/169286424-5cb160c6-4a8d-4342-ad23-b61d5da51b72.png)

Currently figuring out could use this page further in relation to #268 

SETTINGS
--
The Settings page is in an early design phase and should feature a block where the updates can be accessed: 

![Stereum Update](https://user-images.githubusercontent.com/82385103/169274354-1320d9c3-2ced-4ddd-930e-1807e58d3ba8.png)


> Short Explanation

LEFT TOP BUTTON ->  Opens CONFIRMATION dialog box (asked to choose version (?)) -> Initiates Download 
LEFT BOTTOM BUTTON -> Initiate Search for Updates Now. 
Installed Stereum Version -> shows current version
[...]

PR #561, opened 2022-07-19
ADD: running updates one by one

NeoPlays:
- updates can now be run one by one
- available updates are now also shown on Node Page
- added mainnet support


NeoPlays:
one step in the direction of #269

maxbehzadi82 (approved):
gj

PR #578, opened 2022-07-22
ADD: auto reconnect after connection loss

### LINKED ISSUE #560: Deal with connection interrupts
If the connectivity isn't stable to the host it's possible that data is shown/hidden all the time. A better approach is to not refresh the state if the connection dropped and wait for reconnect.

NeoPlays:
closes #560 
and some smaller problems


mabasian (approved):
lg

Issue #588, opened 2022-07-26
updating twice in a row makes the launcher stuck


stefa2k:
First update was correct, rebootet the host also. After the first update I started the update again but it got stuck (no update in the background, but "updating..." didn't disapear anymore on the launcher.

NeoPlays:
this is caused by errors during restarting

PR #589, opened 2022-07-26
REFACTOR: monitoring on Control-Page

### LINKED ISSUE #525: Own SSH Connection for Monitoring
Just like with TaskManager i would open an own SSH connection (maybe an own class) for the monitoring stuff on Control page. This will realease some load of the main connection and ensures that important calls are always coming through.

NeoPlays:
fixes #525 
it worked 10 times in a row but idk why #588 (just to have a reference)
Stereum version is now pulled on setup (if api fails or IS_DEV=true: main HEAD )
also fixed some smaller problems

### DIFF
--- a/launcher/src/backend/Monitoring.js
+++ b/launcher/src/backend/Monitoring.js
@@ -52,4 +52,33 @@ export class Monitoring {
         }
         return []
     }
+
+    //serverNmae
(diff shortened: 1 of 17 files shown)


maxbehzadi82 (approved):
;)

PR #603, opened 2022-08-03
ADD: key import error handling

### LINKED ISSUE #593: Validator account importing with wrong password no error
image
After trying to import a validator account key with the wrong password, it looks the import worked. There should be an error message.

### LINKED ISSUE #594: Old validator account key still in the import list
image
After importing a validator account key, then pressing import again and selecting the same file again, the file appears 2 times in the list.

NeoPlays:
fixes #594 
fixes #593 
there is not a pop up that shows the status of the import:
when there is no error it says: "<number of keys> Keys were imported!"
if there is it prints the error

@maxbehzadi82 you may need to make some ui changes i just copied the delete modal from `FormSetup.vue` and changed it up a little bit 😅

### DIFF
--- a/launcher/src/backend/ValidatorAccountManager.js
+++ b/launcher/src/backend/ValidatorAccountManager.js
@@ -101,13 +101,20 @@ export class ValidatorAccountManager {
 
         }
         try {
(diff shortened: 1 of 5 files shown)


maxbehzadi82 (approved):
lg

PR #604, opened 2022-08-04
UPDATE: Integration tests, replace Math.random with crypto

NeoPlays:
integration tests wont get stuck (will still fail except of Teku)
replaced math.random with crypto (more random)

### DIFF
--- a/launcher/src/backend/ethereum-services/TekuBeaconService.int.js
+++ b/launcher/src/backend/ethereum-services/TekuBeaconService.int.js
@@ -17,7 +17,7 @@ test('teku validator import', async () => {
         name: 'Teku--integration-test--ubuntu-2204',
         image: 'ubuntu-22.04',
         location: 'fsn1',
-        server_type: 'cx31',
+        server_type: 'cxp21',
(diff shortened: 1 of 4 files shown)


launcher/src/backend/ethereum-services/TekuBeaconService.int.js
```diff
@@ -15,7 +17,7 @@ test('teku validator import', async () => {
         name: 'Teku--integration-test--ubuntu-2204',
         image: 'ubuntu-22.04',
         location: 'fsn1',
-        server_type: 'cpx21',
+        server_type: 'cx31',
```
gbayasgalan, line 20: is it necessary to upgrade VM? to run Teku

NeoPlays: ups no hetzner just had issues `Error: resources not available` and then i had to switch server type 😅

gbayasgalan: got it 👍

PR #622, opened 2022-08-05
ADD: Fee Recipient via API

NeoPlays:
waiting for implementation on client side

- [x] [Lighthouse](https://github.com/sigp/lighthouse/issues/3173)
- [x] [Prysm](https://github.com/prysmaticlabs/prysm/pull/10850)
- [x] [Nimbus](https://github.com/status-im/nimbus-eth2/pull/3864)
- [x] [Teku](https://github.com/ConsenSys/teku/issues/6050)


maxbehzadi82:
@NeoPlays which part of the fee recipient r u working on? 
Screenshot 2022-08-06 at 20 16 05

NeoPlays:
> @NeoPlays which part of the fee recipient r u working on? Screenshot 2022-08-06 at 20 16 05

functionality of this thing but the teku client hasn't implemented it yet.. actually it's done already 😅

NeoPlays:
i took out the frontend implementation to avoid future conflicts.
I tested the import on:

- [ ] Teku - Response: ""
- [ ] Lighthouse - Response: null
- [ ] Nimbus - Response: ""
- [ ] Prysm - Response: ""

the imports where successful for all clients where i was able to check the fee recipient in a file
i couldn't find any file for prysm but the api request didn't fail either
I also tested restarting the containers to check if the `--validators-proposer-default-fee-recipient` would overright the api import and they don't so the api has higher priority just like some docs state

stefa2k:
lgtm, but somebody with JS experience should look over it too

mabasian (approved):
lg

PR #665, opened 2022-08-17
FIX: fixed some issues

### LINKED ISSUE #657: Disable "custom installation" and "config import" on Welcome screen
Some useres might not run the one click installation, which could leave the node in a tangling state with no use. Please disable the 2 not working buttons on the welcome screen

maxbehzadi82:
fixed #657


mabasian (approved):
good job

PR #680, opened 2022-08-18
FIX: remove fixed geth tests

stefa2k:
Because it's not possible anymore to run a remote geth (jwt token required) this PR removes all dependencies to remote geth servers.


gbayasgalan (approved):
lg 👍

PR #694, opened 2022-08-22
FIX: issues

### LINKED ISSUE #608: Disable Confirm Button in Plug-In Settings when no changes were made
When there are no changes that have been made in the Plug-In Settings, the confirmation button should be disabled

### LINKED ISSUE #662: Stereum - Notifiation / Support & Log Out
1) Notification Center
2) Support Center
3) Log Out Model

### LINKED ISSUE #671: Task Manager opens up when changing the page
When I change to Control, Node, Staking, everytime the task manager pops up.

@daGscheid do we have some rules on when the task manager shows up on it's own?

### LINKED ISSUE #676: Disable buttons after click
After clicking on buttons they don't get disabled, leading to issues like race-conditions when double-clicking.

All buttons should disable after a single click until the action is reasonable to execute again by the user.

### LINKED ISSUE #689: Status of services break when selecting one for start/stop
When I click to stop or start SSV Network plugin the status of the Teku validator plugin also spinns although there is no start/stop of the Teku going on.

image

Looks like just a display glitch to me.

maxbehzadi82:
fixed #689
fixed #676
fixed #671
fixed #662 
fixed #608


mabasian (approved):
lg

PR #698, opened 2022-08-24
ADD:Validator Individual Options

maxbehzadi82:
single key Grafiti box 
single key exit chain
single key remove
stereum current version


mabasian (approved):
LG

PR #703, opened 2022-08-24
FIX: update & setting issues

maxbehzadi82:
Update panel Stereum btn fixed
update panel ALL btn fixed
Github link changed for support modal
active Confirm btn after any changes in setting


daverolo (approved):
LGTM

PR #706, opened 2022-08-26
FIX: correct sudo permissions and logline argument

### LINKED ISSUE #655: Check if user has correct sudo permissions
Right now there is no check for the correct permissions of a sudo user. Users should be notified if their sudo user don't have the correct rights (`ALL=(ALL) NOPASSWD: ALL` is necessary right now).

There's already a script we can use thanks to @daverolo: https://github.com/daverolo/cansudo

I'd suggest to do this check before checking for OS compatibility and notify the user and prevent installation/login.

daverolo:
Notify the user and prevent installation/login if sudo permissions are missing. (fixes #655)
Emulate method overloading for logline argument.

### DIFF
--- a/launcher/src/backend/NodeConnection.js
+++ b/launcher/src/backend/NodeConnection.js
@@ -29,7 +29,9 @@ export class NodeConnection {
    * identify the operating system of the connected node
    */
   async findOS() {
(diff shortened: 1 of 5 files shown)


mabasian (approved):
LG

PR #710, opened 2022-08-28
check for correct sudo permissions and tiny tweaks

### LINKED ISSUE #655: Check if user has correct sudo permissions

daverolo:
fixes #655


mabasian (approved):
lg

PR #718, opened 2022-08-30
FIX: correct permissions of (dir of) engine.jwt

### LINKED ISSUE #597: The Merge: JWT docker volume creates directory
If a CL docker container boots up before the JWT creation, docker creates a directory and this causes errors on EL side and CL side. If this directory was created ansible throws following error on JWT creation:
```
TIME: 1659087280809773656
PLAYBOOK: /opt/stereum/ansible/controls/genericPlaybook.yaml
TASK: Generate JWT (execution client only)
ACTION: copy
CATEGORY: FAILED
DATA: {"diff": [], "msg": "can not use content with a dir as dest", "_ansible_no_log": null, "changed": false}
```

stefa2k:
fixes #597

### DIFF
--- a/controls/roles/manage-service/tasks/main.yml
+++ b/controls/roles/manage-service/tasks/main.yml
@@ -49,8 +49,8 @@
       state: directory
       owner: "{{ stereum_service_configuration.user }}"
       group: "{{ stereum_service_configuration.user }}"
-      mode: 0700
-      recurse: yes
+      mode: 0777
+      recurse: no
     changed_when: false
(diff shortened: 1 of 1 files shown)


NeoPlays (commented):
I've put a lot of thoughts into this issue and i think that as long as the volumes are handled right this implementation is fine 👍🏼
Maybe @gbayasgalan sees something i dont 😉

gbayasgalan (approved):
looks good to me 👍

PR #728, opened 2022-09-01
FIX: failing test because of not set env

NeoPlays:
Previously Jests Integration tests needed to be run with an env variable (`IS_DEV=true`).
Now the tests can be run without any envs (except HCLOUD_TOKEN).


daverolo (approved):
LGTM

Issue #730, opened 2022-09-01
BUG: updating a service makes the update buttons stuck in disabled


stefa2k:
**Describe the bug**
When updating a service the update buttons stay disabled even after the update finishes

**To Reproduce**
1. Click to update a service
2. wait until the update finishes

**Expected behavior**
Update buttons get enabled again after the update finishes

**Screenshots**
![image](https://user-images.githubusercontent.com/54934211/187912697-f1c59849-bee5-46c2-b0b3-44c9fe3be701.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Launcher version `2.0-rc2`

**Node (please complete the following information):**
 - OS: Ubuntu 22.04
 - Affected service: Updates

**Additional context**

stefa2k:
Also it looks like it prevents the Expert Mode data from getting refreshed. Update was finished but the Expert Mode still showed the old version in it. After restarting the Launcher the service's version was correct in the Expert Mode.

PR #733, opened 2022-09-02
FIX: update btns getting stuck

NeoPlays:
#730 
I'm pretty sure this bug is caused by errors interrupting the refresh flow.
these changes should prevent this from happening.
But i wasn't able to reproduce this bug so i wont close this issue yet.


mabasian (approved):
LG

PR #737, opened 2022-09-06
FIX: teku metrics to default (all categories)

stefa2k:
https://docs.teku.consensys.net/en/latest/Reference/CLI/CLI-Syntax/#metrics-categories


NeoPlays (approved):
lgtm 👍🏼

PR #738, opened 2022-09-06
ADD: change graffiti

NeoPlays:
Changing graffitis for all keys for Lighthouse, Prysm and Teku Nodes is now possible. 
On Nimbus nodes the only way to change the graffiti while the service is running, is via their rest api.
Doing this via ansible and curl would slow things down a lot, so i left it open for now

### DIFF
--- a/launcher/src/backend/ValidatorAccountManager.js
+++ b/launcher/src/backend/ValidatorAccountManager.js
@@ -230,39 +230,45 @@ export class ValidatorAccountManager {
 
     async setGraffitis(graffiti){
         let services = await this.serviceManager.readServiceConfigurations()
-        let client = undefined
-        let possibilities = ['Validator', 'Beacon']
-        let i = 0
-        while (client === undefined && i <= 1) {
-            client = (services.find(service => service.service.includes(possibilities[i])))
-            i++
-        }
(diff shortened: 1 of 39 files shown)


stefa2k:
This update means we need to change the existing services, right?

NeoPlays:
> This update means we need to change the existing services, right?

yes for the graffiti file attribute

stefa2k:
Okay, that's still to do then, same with #737

launcher/src/backend/ValidatorAccountManager.js
```diff
@@ -228,4 +228,42 @@ export class ValidatorAccountManager {
         return URL
     }
 
+    async setGraffitis(graffiti){
```
stefa2k, line 231: it's possible to run mulitple validator services

stefa2k (changes requested):

launcher/src/backend/ValidatorAccountManager.js
NeoPlays: done 👍🏼

stefa2k (changes requested):
There should be a verify of the validator services when started creating a `graffiti.yaml` in the molecule tests.

stefa2k (approved):
Nice work! 🚀

PR #740, opened 2022-09-06
ADD: launcher updates

NeoPlays:
This adds launcher updates in the background without any dialogs atm.
Updater-Activities are logged via `electron-log`


NeoPlays:
> What happens when the update fails? Is the app still able to run?

Yes, the updater finds a realease to update and downloads it. So if that fails nothing happens.
and the installation after restart shouldn't be problem if the steps before succeed.
The updater has no issues finding the right repo and the yamls are also accessable as long as they exist.

maxbehzadi82 (approved):
Jetzt gehts

PR #742, opened 2022-09-07
ADD: (tests for) config update #738 #737

stefa2k:
this updates existing service configurations to be fully compatible with #737 and #738


NeoPlays (approved):
lgtm 🚀

Issue #601, opened 2022-08-01
Synchronisation - Wiring 


daGscheid:
For reference
![image](https://user-images.githubusercontent.com/82385103/182247845-1bd00736-21b4-4f46-826b-89c6a89420a9.png)

Wire the current status of synchronisation of by the node's installed EXECUTION & CONSENSUS clients with the following status changes:

### ARROW CIRCLE (!!!)

Green = ALL CLIENTS Up to date
![Synchronisation_Icon_Synchronized](https://user-images.githubusercontent.com/82385103/182248932-32de565d-6658-4968-b4b7-9dad8254ed52.gif)

Blue = AT LEAST ONE client is synchronizing
![Synchronisation_Icon_Active](https://user-images.githubusercontent.com/82385103/182248928-13b2d1cb-eeaf-43e0-9d15-6df20acf989e.gif)

RED = ERROR in SYNCH
![Synchronisation_Icon_Error](https://user-images.githubusercontent.com/82385103/182248787-40fd6d5c-6ab7-4ef7-9dac-6abe9f17fdfb.gif)

### FONT COLOR (!!!)

Font should be different colour depending on sync status of the client. 

GREEN FONT = Client is up to date

BLUE FONT = Client is syncing

RED FONT = Client is not syncing

If you have question reach out to me @mabasian

daGscheid:
@gbayasgalan we had this in Stereum 1.0, do you know how we did this?

NeoPlays:
> @gbayasgalan we had this in Stereum 1.0, do you know how we did this?

I can take a look at it currently waiting for fee recipient ui 😅

gbayasgalan:
> @gbayasgalan we had this in Stereum 1.0, do you know how we did this?
beacon-APIs used to display syncing status. API-link is here: https://ethereum.github.io/beacon-APIs/#/Node/getPeerCount. Also take look at `refreshPeersModel` function on`stereum1` https://github.com/stereum-dev/ethereum2-control-center-web/blob/main/frontend/src/components/cc/ServicesOverview.vue#L397.

NeoPlays:
> > @gbayasgalan we had this in Stereum 1.0, do you know how we did this?
[...]

PR #753, opened 2022-09-12
ADD: nodestats data syncstatus and p2pstatus

daverolo:
Data for wiring syncstatus (#601) and p2pstatus on controls page


NeoPlays:
`curl` needs to be installed on setup

PR #758, opened 2022-09-13
ADD: updating to HEAD

NeoPlays:
By setting the env var `IS_DEV=true`
Stereum-Controls-Updates are always updating to HEAD of main


stefa2k (approved):
lgtm

PR #759, opened 2022-09-13
FIX: Bugs on staking page

NeoPlays:
Bug 1: Keys that are in the process of being deleted weren't marked red ✅ 
Bug 2: Beaconcha.in returns data of `null` instead of `[]` for invalid (not deposited) keys which led to an error ✅


maxbehzadi82 (approved):
hero

PR #762, opened 2022-09-14
ADJUST: Integration Tests

NeoPlays:
Add a loop for retrieving logs so tests don't fail because of not long enough logs.


stefa2k (approved):
Maybe this should be an abstract function to use in each test, not copy & paste code :)

Good enough for now

NeoPlays:
> Maybe this should be an abstract function to use in each test, not copy & paste code :)
> 
> Good enough for now

yeah you're right 😅

PR #763, opened 2022-09-14
ADD: restart-services role & github actions molecule schedule

stefa2k:


gbayasgalan (approved):
lg

launcher/src/backend/NodeConnection.js
```diff
@@ -857,13 +860,42 @@ export class NodeConnection {
...
+      let after = await this.getTimeStamp()
+      return parseInt(after) - parseInt(before)
     } catch (err) {
       log.error("Error occurred running stereum updates:\n", err)
     }
   }
 
+  async getTimeStamp(){
+    try {
+      let val = await this.sshService.exec('date +%s')
```
stefa2k, line 875: You could just take the time diff off of the local desktop. Would save a round to ssh command.

NeoPlays: I was a bit in a hurry doing this 😅
Improving it as soon as i'm back home 👍🏼

NeoPlays:
Services restart only after update when `autoupdate=true` is included in the config.
This is not yet implemented so i decided to take the timestamp after service updates so that updated services also get rebooted.

NeoPlays (approved):

PR #769, opened 2022-09-15
ADD:Custom installation

maxbehzadi82:
Add Panel,
Remove Service
Modify Panel,
Replace Services
502, 501, 500, 499


daverolo (approved):
LGTM

PR #772, opened 2022-09-15
REFACTOR: wiring tweaks

daverolo:
Changed nodestats response structure.
Each data element has now its own return code/info/data fields to improve error handling.
Adjusted icon and lower/higher output in syncstatus component


mabasian (approved):
LG

PR #774, opened 2022-09-15
FIX: storage status and and geth labels

daverolo:
Provides storage status for all services.
GETH labels for Prometheus in getSyncStatus now verified.


mabasian (approved):
Good job🚀

PR #786, opened 2022-09-20
FIX: Custom installation some functionality

maxbehzadi82:
The selected service has a border.
Cancel the latest selection after click on another.


mabasian (approved):
Very clean codes, Good job 👍🏼

Issue #785, opened 2022-09-19
BUG: sync status component shows error on client initialisation


daverolo:
**Describe the bug**
The [sync status component](https://user-images.githubusercontent.com/82385103/182247845-1bd00736-21b4-4f46-826b-89c6a89420a9.png) is showing [sync error](https://user-images.githubusercontent.com/82385103/182248787-40fd6d5c-6ab7-4ef7-9dac-6abe9f17fdfb.gif) while the node is initializing. This happens <strike>mostly when using Prysm</strike> with all clients.
From a technical point of view the current implementation is correct because Prometheus data is not available for a few minutes during initialisation, however from a human perspective it feels wrong because the services are up and the clients are in fact already synchronizing in the background.

**To Reproduce**
Setup a new node with default installation using PRYSM and GETH.

**Expected behavior**
A information that the node is currenlty initializing, for example a [synchronization cycle](https://user-images.githubusercontent.com/82385103/182248928-13b2d1cb-eeaf-43e0-9d15-6df20acf989e.gif) in orange color.
If the sync data is still not available in Prometheus after _X_ minutes, [show an error](https://user-images.githubusercontent.com/82385103/182248787-40fd6d5c-6ab7-4ef7-9dac-6abe9f17fdfb.gif) as usual.
Also pay attention that there should **not** shown an error when the data is there but the lower slot is zero **during initialization** (see second screenshot with teku) except the second client (in that case geth) has still no data.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/107847185/191070996-7fc6379a-ef32-4b5a-8928-9dba95ca9ffb.png)
![grafik](https://user-images.githubusercontent.com/107847185/191074236-ce765a7c-ee5a-43ba-be71-e8a1d56fe48a.png)

daGscheid:
So what you mean - I will design a "default" one - in yellow (mirroring the one used for the resync option)

daGscheid:
![Synchronisation_Icon_Connecting](https://user-images.githubusercontent.com/82385103/191621819-82bd5b7f-e35c-4e0d-89aa-c0ae44801f3f.gif)


I made this one. LMK what you think.

 Should be the "default" one if you guys get what I mean. I think currently it is the green one, and this is damn confusing haha

daverolo:
> I made this one. LMK what you think.

I think that icon hits the topic just perfect.

PR #788, opened 2022-09-20
FIX: improve query/filter for prysm sync status

daverolo:
Fixes broken sync status widget when Prysm was used.
Improves query/filter for all clients of the sync status widget.
This is explicitly **not** the solution for #785!


mabasian (approved):
Good job!

PR #801, opened 2022-09-26
ADD:select validator service for the keys

maxbehzadi82:
Add Picking a service for single or a multiple of Validator keys.
Add switching network on manage node and display on the node page
Fix some custom install issues


NeoPlays (approved):
lgtm 🚀

PR #807, opened 2022-09-28
FIX: d&d and some other issues

maxbehzadi82:
Drag & drop fixed
Tutorials disabled those are not exist
task manager some fixes


NeoPlays (approved):
🔥

PR #809, opened 2022-09-30
REFACTOR: rework syncstatus widget

daverolo:
Font color per client and adjusted icon in syncstatus widget.
Queued requests for control components to avoid stuttering.
Some tiny fixes and improvements.


launcher/src/pages/TheControl.vue
```diff
@@ -40,63 +40,81 @@ export default {
     }),
   },
   mounted() {
-    this.polling = setInterval(this.refresh, 1000); //refresh services
+    this.polling = setInterval(this.refresh, 100); //refresh services
```
NeoPlays, line 43: a bit too fast don't you think so ? 😅

daverolo: I think it's fine because there is no request executed at all since requests are queued, see TheControl.vue -> requestQueued(). The polling ist just fast to make a request as soon as the queue is free. In the requestQueued() method usually only vars are checked, which is IMHO a standard operation in js for a 100ms timeout.

NeoPlays: I'm just thinking if we run into some weird problems and we catch an error over and over again, the console is spammed with 10 messages per second 😅 
Another thing are local nodes where these commands take like no time this results in heavy load or am i wrong?

I'm just thinking if it doesn't need to be that fast don't make it that fast 😅

NeoPlays (approved):
lgtm

PR #820, opened 2022-10-05
FIX: remove check for sudo group

daverolo:
It is not needed to check for the sudo group.
Thanks to Stereum Discord user redtux for investigating that issue.


mabasian (approved):
Good job!

PR #843, opened 2022-10-20
ADD: mevboost - oneClick installation

gbayasgalan:
part of "one Click install - mevboost"

- at the moment, Mevboost takes only static relays from mevboost.js (one for mainnet, other for testnet -> goerli) 
- tested with all CL

to do: 
 - dynamic relay(s) from UI - for now relays not wired with UI 
 - add mevboost in dependencies


maxbehzadi82 (approved):
lgtm

PR #869, opened 2022-10-31
FIX: OneClickInstallation and Client Tests

NeoPlays:
MevBoost integration caused some issues with the OneClickInstallation


gbayasgalan (approved):
lg 👍

PR #870, opened 2022-11-02
ADD: mevboost dep script

NeoPlays:
error prone for missing dependencies,
update-changes script for mevboost dependency


daverolo (approved):
LGTM

PR #904, opened 2022-11-11
FIX: molecule test validator-delete-API

gbayasgalan:
- regexp is used instead of `jmespath`
- update: teku - https://github.com/ConsenSys/teku/releases/tag/22.10.2 (bug fix)


NeoPlays (approved):
lgtm

PR #907, opened 2022-11-11
FIX: apt update on install

stefa2k:
sometimes the `add-apt-repository` command on install doesn't update the apt repositories afterwards, causing ansible not get installed properly


stefa2k:
doesn't fix it

PR #914, opened 2022-11-11
FIX: bugs before release 

NeoPlays:
- Installation paths where normalized based on operating system and not linux
- when connecting mevboost to new not installed yet service, task failed
- reset stored stereum settings on logout


maxbehzadi82 (approved):
lgtm

Issue #650, opened 2022-08-12
Non Default SSH Port Support


NeoPlays:
Stereum 2 only supports port 22 atm.
The js backend supports the option for another port and is integrated to some point.
One thing missing is some Input on the Login Page for the port.

stefa2k:
Please also save the port to the launcher's config. Right now it doesn't contain the port number:
```
			{
				"name": "test2",
				"host": "2.3.4.5",
				"user": "deployyyy",
				"keylocation": "",
				"useAuthKey": true
			},
```

CosmiCloud:
On the login page you could just allow the user to specify the port they want by using : after the IP.

pepawel:
If for some reason you can't or don't want to change SSH port, you could use port forwarding **on your local machine** as a workaround.

There are many possibilities, personally I use [socat](http://www.dest-unreach.org/socat/) which is an easy to install package in many Linux distributions.
Run this command on your local machine replacing HOST and PORT with address and ssh port of the machine you would like to connect to:
`$ sudo socat -dd TCP4-LISTEN:22,fork TCP:HOST:PORT`

Then configure Stereum to connect to localhost.

Socat works on UNIX, so there is macOS version. I've also found unofficial build for Windows.

PR #948, opened 2022-11-30
Notif handling

mabasian:


mabasian:
UI for the ssh port also is ready 
Fix #650

PR #971, opened 2022-12-16
ADD: Condition for DisabledStaking.vue

### LINKED ISSUE #965: Staking Page - No Validator Client
> To better track this. 
We should disable / deactivate components on the staking page, whenever:

[ ] there is no Validator Client installed.
EN - STAKING TAB – 1

[ ] the validator client is turned off.  
EN - STAKING TAB – 2

Image to fill placeholder is currently in works & will be provided at a later date.

NeoPlays:
Checking if there is a validator client and if it is running
closes #965


gbayasgalan (approved):
Lg

PR #974, opened 2022-12-21
REFACTOR: prevent nodatalayer on first load in p2p- and syncstatus widget

daverolo:
- Made sure that the NoDataLayer is only shown after the widget has at least one time successfully initialized
- Cleaned dead code from SyncStatus Widget


mabasian (approved):
LG 🚀

PR #978, opened 2022-12-22
ADD: Restart after initial Setup

### LINKED ISSUE #413: Restart after initial setup
At the moment the setup process also installs OS updates. Some of these updates might require a restart of the host to take effect.

Please look up when a restart is necessary and restart the host properly.

NeoPlays:
This checks for a requireed restart and restarts accordingly.
Optimised refresh loops to prevent errors while connection cuts when the node restarts.
closes #413


gbayasgalan (approved):
Lg

PR #982, opened 2022-12-22
FIX: Erigon Config

NeoPlays:
Erigon changed their default behaviour:

### Embedded Consensus Layer

By default the Engine API is disabled in favour of Erigon native Embedded Consensus Layer, if you want to either stake or sync an external Consensus Layer, run Erigon with flag `--externalcl`.


gbayasgalan (approved):
lgtm 👍

PR #989, opened 2023-01-02
ADD: SSV Expert Mode

### LINKED ISSUE #691: SSV Network expert mode
SSV Network also got an additional config file, the `config.yaml` which should also be editable like the service's yaml.

NeoPlays:
### SSV Expert Mode
This adds a text-area in the service config window for the `SSVNetworkService` which is used to configure the ssv config. fixes #691 

### Config Window Adjustments
other options are now invisible when a text-area is open to make it as big as possible

### Fix `SSVNetworkService`
due to merge adapations the network written in the ssv `config.yaml` was `goerli` which ssv does not translate to `prater`.
Hard coded it to `prater` for now


gbayasgalan (approved):
lgtm 👍

PR #1015, opened 2023-01-17
FIX: Arrow-Buttons on 'Edit Node' for services don't cycle

PatrickRL:
Fixed by adding a function to cycle backward

### DIFF
--- a/launcher/src/components/UI/node-manage/SidebarManage.vue
+++ b/launcher/src/components/UI/node-manage/SidebarManage.vue
@@ -20,9 +20,9 @@
     <div :class="{ 'run-sidebar': showSidebar }" class="manage-sidebar">
       <div class="categoryBox">
         <div class="category">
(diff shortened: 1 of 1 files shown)


launcher/src/components/UI/node-manage/SidebarManage.vue
```diff
@@ -150,6 +150,41 @@ export default {
           break;
       }
     },
+    selectCategoryTitlePrev() {
```
NeoPlays, line 153: this will work but in software development is `KISS` -> `Keep It Small & Simple` the key

you made an almost exact copy of an function so basically you have now lines in your code that have the exact same perpuse
you can pass variables like this: `@click="selectCategoryTitle("right")"` for example
try adjusting the existing function
you can `ADD` -> `COMMIT` -> `PUSH` on this branch and it will update the PR as well

NeoPlays (changes requested):

NeoPlays (approved):

Issue #852, opened 2022-10-21
BUG: Provide valid Sync Status for Nethermind EC


daverolo:
**Describe the bug**
Sync Status for Nethermind EC is invalid.

**To Reproduce**
1. Setup a fresh node with Nethermind EC
2. Select the "Control" Tab
3. See invalid values for Nethermind in "Sync Status" Widget

**Expected behavior**
A valid representation of current Nethermind EC Sync Status

**Screenshots**
![grafik](https://user-images.githubusercontent.com/107847185/197149303-617c7bd6-fdcf-49a7-ad58-aac5c77a41c0.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Launcher version: 2.0.0-RC.6

**Node (please complete the following information):**
 - OS: Ubuntu 22.04
 - Affected service: Nethermind EC v1.14.3

**Additional context**
Nethermind EC does not provide data for Prometheus at all to retrieve a valid Sync Status.
The currently tested labels `['nethermind_blocks','nethermind_blocks_sealed']` are wrong.
An option for the future could be to request the Sycn Status via RPC-API (untested) .

stefa2k:
There's some more data available of nethermind:
- https://github.com/NethermindEth/nethermind/pull/4887
- https://github.com/NethermindEth/nethermind/issues/4860
- https://github.com/NethermindEth/nethermind/issues/4842

OlegJakushkin:
Dear @daverolo:
Currently syncing with snap sync NM+Prysm. Metric you have referenced works.
![image](https://user-images.githubusercontent.com/2915361/206455910-87d1edad-05a9-4d7e-a733-e8e572e13e01.png)

@daverolo Could you please compare the arguments you have used to run it?

Here are mine:
```
.\Nethermind.Runner.exe -c mainnet -dd C:/work/ethdata --Metrics.Enabled true --Metrics.ExposePort=9091 --JsonRpc.Enabled true --JsonRpc.Port 8547 --Sync.SnapSync true --JsonRpc.JwtSecretFile C:/ethereum/consensus/jwt.hex

.\beacon-chain-v3.1.2-windows-amd64.exe --mainnet --execution-endpoint=http://localhost:8551/ --jwt-secret=C:/ethereum/consensus/jwt.hex --checkpoint-sync-url=https://beaconstate.info/ --genesis-beacon-api-url=https://beaconstate.info/ --datadir=C:/work/ethdata
```

with 

[...]

PR #1018, opened 2023-01-17
FIX: provide solution for nethermind issue 852

daverolo:
this is a partial solution for #852 (@NeoPlays -> currently missing open ports on existing nodes with ansible and custom configuration)


NeoPlays:
i will commit right to this branch so we got everything together

NeoPlays (approved):
🚀

PR #1021, opened 2023-01-18
ADD: gnosis basic validator stats

NeoPlays:
This adds basic validator stats on gnosis fetched very epoch from [https://beacon.gnosischain.com/](https://beacon.gnosischain.com/)


daverolo (approved):
LGTM

PR #1033, opened 2023-01-24
UPDATE: Nethermind, qol changes

NeoPlays:
## Nethermind -> `1.16.1`
new gnosis nodes will use `gnosis` instead of `xdai`

## Failing Lodestar tests
mistake in the service config

## Choose Path of Largest Volume
if the path of the largest volume was the root dir the install directory ended up with `/stereum`
changed to `/opt/stereum` when path of largest volume equals `/`.

## Custom Installation
chooses path of largest volume when installing via custom installation

## MEV Boost OneClickInstaller
the right image version is now queried

## Prometheus Node Exporter
readded update-changes script and tests for next release


gbayasgalan (approved):
lg

PR #1048, opened 2023-01-31
FIX: changing ports, service doc-links

NeoPlays:
## Changing Ports
When installing Flashbots-MEV-Boost via custom installation, the ports of the selected consensus client were changed without any duplicated binds.

## Service Docs
the button in expert mode now opens the specific docs in the browser


mabasian (approved):
LG

PR #1051, opened 2023-02-01
FIX: delete services on custom installation paths when nuking node

PatrickRL:


NeoPlays (changes requested):
As you might have seen there are `this.taskManager.otherSubTasks.push` which add a task in the taskmanger you might want to implement this too.
Also if you look at the other commands executed there is one that removes the install directory so basically you dont have to remove volumes that are in the install directory.

NeoPlays:
as you might have seen for the task status we use the rc code of that command. you will have more ssh calls so also more rc codes so you'll have to check if one of the rc codes is not 0 and if so then set the status as failed.
example:
```
    this.taskManager.otherSubTasks.push({
      name: "remove Volume Directories",
      otherRunRef: ref,
      status: (await MyAwesomeFunctionThatReturnsTrueorFalse),
    });
```

launcher/src/backend/NodeConnection.js
```diff
@@ -3,6 +3,7 @@ import { StringUtils } from "./StringUtils";
 import { NodeConnectionParams } from "./NodeConnectionParams";
 import { nodeOS } from "./NodeOS";
 import { ServiceVolume } from "./ethereum-services/ServiceVolume";
+import { ServiceManager } from "./ServiceManager";
```
NeoPlays, line 6: This Import is not necessary

launcher/src/backend/NodeConnection.js
```diff
@@ -777,10 +777,27 @@ export class NodeConnection {
...
+    for(let service of serviceConfigs) {
+      let sID = service.id;
+      for(let path of service.volumes) {
+        let sPath = path.destinationPath.substring(0, path.destinationPath.indexOf(sID)) + sID + "/";
+        if(sPath.includes(sID) && !sPath.includes("/etc/stereum") && sPath != sPathCheck){
+          sPathCheck = sPath;
+          this.taskManager.otherSubTasks.push({
+            name: "remove Volume Directories",
+            otherRunRef: ref,
+            status: !(await this.sshService.exec("rm -rf " + sPath)).rc,
```
NeoPlays, line 795: this would make a subtask for every volume which is fine if you give them different names. in this case they're all named `remove Volume Directories`

NeoPlays (approved):
looks good 🚀

PR #1062, opened 2023-02-07
FIX: tunnel issue

NeoPlays:
## Tunnel Issue
reloading the app in dev mode, caused the app to throw an error due to `tunnel-ssh` configured with `autoClose: true` some tunnels where closed some not and at the same time the app was already looking for new free ports to open which then caused this issue


mabasian (approved):
LGTM

PR #1065, opened 2023-02-08
Fix tooltips debug for Mevboost

mabasian:
Resolved issues with the display of tooltips in Mevboost by thoroughly checking the code and implementing necessary fixes. The tooltips now display correctly and provide accurate information to users.


gbayasgalan (approved):
lg

PR #1069, opened 2023-02-10
Added pruning setting to Nimbus

PatrickRL:
Added pruning to Nimbus, with toggle option between prune and history setting


launcher/src/backend/ethereum-services/NimbusBeaconService.js
```diff
@@ -56,6 +56,7 @@ export class NimbusBeaconService extends NodeService {
         '--keymanager-address=0.0.0.0',
         '--keymanager-token-file=/opt/app/validators/api-token.txt',
         '--jwt-secret=/engine.jwt',
+        '--history=Prune',
```
NeoPlays, line 59: `It is possible to switch between prune and archive modes.`
pls keep the settings lower-case

NeoPlays (changes requested):
in `services.js` as well

NeoPlays (approved):

PR #1075, opened 2023-02-14
Add Log Navigation and Refactor Log Plugin Header Style

mabasian:
This pull request adds a new log navigation feature inside the log plugin and refactors the header section's CSS styles. The log navigation provides users with an easy way to navigate between different logs, while the header style refactor simplifies and improves the readability of the code. These changes should make the log plugin's header section more consistent with the rest of the plugin and easier to maintain.


daverolo (approved):
LGTM

PR #1100, opened 2023-02-26
Enhancement: fix most of eslint errors and warnings

montajebi:


daverolo:
i don't know if this is still work in progress but i would highly suggest a dev meeting in voice channel before merging is performed

PR #1108, opened 2023-03-01
FIX: SSV Modal

NeoPlays:
This fixes the current issues with on the ssv modal


PatrickRL (approved):
Errors I've previously gotten are now gone and it seems to be working correctly now.

Issue #1037, opened 2023-01-24
FR: Graphical Enhancement - Nuke Node Command


daGscheid:
**Is your feature request related to a problem? Please describe.**

Whenever you click the Nuke Node Button, it is just gone. We had the idea during Bogota to enhance this experience a bit by adding an animation to confirming the DELETION. Lin and I are currently working on something we hope to present soon. 

**Describe the solution you'd like**

![image](https://user-images.githubusercontent.com/82385103/214364020-f0eeaafb-07da-4ba1-b3df-880afae69b45.png)

When I click this: BADA BANG (no sound haha) & DOPAMINE + BRAIN 

**Additional context**

SOON :)

Mengine:
would be great to be able to see a log of what is actually being deleted also :)

daGscheid:
Made by our artist ProudHands for the occasion. 

![NukeNode](https://user-images.githubusercontent.com/82385103/222114937-427457ae-b78c-48ff-b886-2e14b0b72bda.gif)

We are currently working on the modal. Will still take a while, till we fit the deletion text & a good way to inspect it.  Some work in progress from today: 

https://user-images.githubusercontent.com/82385103/222115270-55e81b7b-6eb4-48d2-acc8-942493cd2e94.mov

PR #1127, opened 2023-03-14
REFACTOR: show log during nuke node process

daverolo:
- Shows main tasks that happen in the background while a node is nuked
- Part of #1037 - requires frontend style adjustments


mabasian (approved):
LGTM🚀

PR #1128, opened 2023-03-14
ADD: Stereum App Notifications and Provisioning

NeoPlays:
## NotificationService
The notification service known from stereum 1 is reimplemented into stereum 2.0. The service automatically installed for all OneClickInstallations. It can also be added via Custom Installation.

## Grafana Alerts
Grafana alerts can now be provisioned in a similar way as the grafana dashboards. These automated steps are checked in the notifications molecule test.

## Stereum Mobile App
Scan the qr code in your launcher with the mobile app to connect your phone to the alerts


daverolo (approved):
LGTM

PR #1161, opened 2023-03-31
Resync clients

PatrickRL:
Added option to resync a service from genesis or perform checkpoint sync

### DIFF
--- a/launcher/src/backend/ServiceManager.js
+++ b/launcher/src/backend/ServiceManager.js
@@ -159,7 +159,7 @@ export class ServiceManager {
       });
   }
 
-  async chooseServiceAction(action, service) {
+  async chooseServiceAction(action, service, data) {
     switch (action) {
       case "pruneGeth":
         if (service.service === "GethService") {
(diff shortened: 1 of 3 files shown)


launcher/src/backend/ServiceManager.js
```diff
@@ -177,14 +177,69 @@ export class ServiceManager {
...
+        vol.servicePath === "/opt/data/geth" ||
+        vol.servicePath === "/opt/data/erigon"
+    ).destinationPath;
+    await this.manageServiceState(client.id, "stopped");
+
+    if (dataDir.length > 0) {
+      await this.deleteDataVolume(dataDir);
+    }
+    //Prysm specific
+    if (client.service == "PrysmBeaconService") {
```
NeoPlays, line 211: You should avoid client specific exceptions. You could have checked if `service.command` is a string. If so -> convert to array and later back to string. This way you can handle all clients the same way and dont need this big `if else block`.

The reason behind this is if there is another client thats `command` is a string, I'd have to add this one here too. Or if Prysm is refactored and has no longer `command` as a string then this doesn't work as well.

NeoPlays (approved):
Looks good otherwise 🚀

PR #1168, opened 2023-04-05
ADD: Option to configure Prometheus Data Retention

PatrickRL:
ADD: option to configure data retention of Prometheus


launcher/src/store/services.js
```diff
@@ -423,9 +423,17 @@ export const useServices = defineStore("services", {
           hIcon: "/img/icon/service-icons/prometheus.png",
           linkUrl: "",
           docsUrl: "https://prometheus.io/docs/introduction/overview/",
-          headerOption: true,
+          headerOption: false,
```
NeoPlays, line 426: Do you need to change this?

NeoPlays (changes requested):

launcher/src/store/services.js
NeoPlays: with this open are you removing the header icon:
grafik

PatrickRL: Accidentally left this change after testing something

NeoPlays (approved):
👍🏼

NeoPlays (approved):

PR #1185, opened 2023-04-13
FIX: Nimbus MevBoost Dependencies

NeoPlays:
When changing the mev config `--payload-builder=true` was always added to nimbus


PatrickRL:
Looks good!

gbayasgalan (approved):

PR #1188, opened 2023-04-16
FIX: gnosis active since calculation

gbayasgalan:
fixing validator state: `active since` on Gnosis 
gnosis Epoch time: 80 sec while goerli and mainnet 384sec


daGscheid (approved):
Looks good

PR #1194, opened 2023-04-19
ADD: Validator State into Staking page

MaxTheGeeek:
- [x] Validator State
- [x] Validator Selector dropdown
- [x] Tabbar to pick services
- [x] New Icons


NeoPlays:
is this ready for review?

daGscheid:
@MaxTheGeeek I just want him to take a look at it to discuss what we should do for multiple 100 keys. 
(we had some talks at EDCON).  
For more than 5 keys, we have to think of something different in the long run. 
I will open an issue for it. 
It looks good so far, but just to write it down - we should test the block production one if we can. We have some questions that are hard to test in dev. 

It usually throws a Zero, we expect it to throw a slot, if this isn't the case: 🤔 

There are also some cases left we have to work for, whenever the Node is not in sync f.e.

daGscheid:
Should have started with this 😅

MaxTheGeeek:
> Should have started with this 😅 Sorry >.>
Should I keep it as a draft?

daGscheid:
> > Should have started with this 😅 Sorry >.>
> > Should I keep it as a draft?

🤷‍♂️

NeoPlays (approved):
I think this is finally done

Issue #1145, opened 2023-03-21
FR: Add Path to Open WS Endpoint SSH Tunnel 


daGscheid:
**Is your feature request related to a problem? Please describe.**
One of our Discord users `definevalue`  asked to not only be able to open an RPC Endpoint SSH Tunnel, but also a WS Endpoint SSH Tunnel via the U.I. . 

**Describe the solution you'd like**
Something similiar to 
![image](https://user-images.githubusercontent.com/82385103/226561751-2dc0a3ae-da10-4da0-8494-5a5deefee6e2.png)
Maybe adding it to the component itself (changing its name) or adding it as additional component to the dashboard (we still got some room) 

**Describe alternatives you've considered**
I don't negotiate with terrorists 
(I fell into funny today, will make me laugh when I forget about it and read it again - what I want to say is: ignore me)

daverolo:
since there is room on bottom (see sscreenshot), i would suggest a separate widget with a name like _RPC-WS_ that has the exact same format as the other two
![grafik](https://user-images.githubusercontent.com/107847185/226889994-d6473cb8-9903-4217-9909-36e66a4fe5d3.png)
i suggesting this solution because each widget is scrollable and can also hold **multiple** clients.

daverolo:
The implementation works for new nodes where the WS port (usually 8546) is opened during installation.
For existing nodes the WS port currently has to be opened manually or the WIdget will load endless.
**This should be fixed via Ansible or the widget must be disabled before the next release!**

daverolo:
Since  #1205 is merged the dashboard is not blocked anymore but the widget will still load endless if the WS port is not opened manually. So it is not critical anymore, however the Ansible fix (see above) is still an open task.
![grafik](https://user-images.githubusercontent.com/107847185/235304726-f8f541e8-b2ce-462f-a48a-6d2fb273f1c8.png)

daverolo:
Also the Ansible fix to open the WS port on existing nodes was added by @NeoPlays with #1222 before RC.18 release.

PR #1205, opened 2023-04-28
FIX: dashboard and object availability

daverolo:
These are a few general fixes to:
- avoid an application crash if an object is not available and
- keep the dashboard available after #1203 was added, respectively aslong #1145 is open (see Ansible requirements)

### Summary
- Checks object availability in SyncStatus view
- Checks object availability in P2PStatus view
- Checks object availability in PortList view
- Checks "rpc" and "ws" object availability in (monitoring) backend


mabasian (approved):
LGTM

PR #1215, opened 2023-05-04
ADD: Recommanded check points

PatrickRL:
Added drop-down with recommended checkpoint urls in OneClick Installer and Resync Modal


daverolo (approved):
LGTM

Note: Teku requires a different URL sufix than all other clients. Added this to the PR

PR #1240, opened 2023-05-19
Feature implement validator stats backend function

montajebi:


NeoPlays (approved):
lgtm 🚀 
@daverolo what do you think?

launcher/src/backend/Monitoring.js
```diff
@@ -2517,6 +2517,131 @@ rm -rf diskoutput
...
+
+    const { genesis_time } = JSON.parse(genesisResShell.stdout).data;
+    const current_time = Math.floor(Date.now() / 1000);
+    const slot_time = 12;
+    const slot_timeout = slot_time - ((current_time - genesis_time) % slot_time);
+    const current_slot = Math.floor((current_time - genesis_time) / slot_time);
+    const current_epoch = Math.floor(current_slot / 32);
+
+    output = { currentEpoch: current_epoch };
+
```
daverolo, line 2543: the last argument below ({"Content-Type": "application/json"}) is not needed since this is the default anyway

daverolo (approved):
code looks good to me 🚀 
added a few comments where defaults could be used.
just for the record: could not test this on a real val since the function is not used anywhere atm

PR #1271, opened 2023-06-12
Support fee recipient for validator accounts

mabasian:


mabasian:
@NeoPlays the feeRecepientConfirmHandler method is ready

PR #1282, opened 2023-06-14
ADD: Client Switch

PatrickRL:
Added option to switch out clients with other clients

### DIFF
--- a/launcher/src/backend/ServiceManager.js
+++ b/launcher/src/backend/ServiceManager.js
@@ -394,7 +394,7 @@ export class ServiceManager {
         break;
       case "Charon":
         filter = (e) => e.buildConsensusClientHttpEndpointUrl();
-        command = "--beacon-node-endpoints="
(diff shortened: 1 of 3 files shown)


launcher/src/backend/ServiceManager.js
```diff
@@ -907,50 +975,68 @@ export class ServiceManager {
     let ELInstalls = tasks.filter((t) => t.service.category === "execution");
     ELInstalls.forEach((t) => {
       let service = this.getService(t.service.service, t.data);
-      t.service.config.serviceID = service.id;
+      if (!Object.hasOwn(t.service, "config")) {
+        t.service.id = service.id;
+      } else {
```
NeoPlays, line 980: Idk exactly what this is for, but this breaks the regular Custom Installation of clients.

(same for CC, VC and Mev)

NeoPlays (changes requested):

Tobee77 (approved):

PR #1317, opened 2023-06-26
FIX: Client restart on switch

PatrickRL:
Clients who's config have been changed through client switching now restart afterwards


NeoPlays (approved):
lgtm

PR #1340, opened 2023-07-05
fix the ports point warning status 

mabasian:
warning alarm shows when RPC, data API and WS points are open in the alarm boxes


daverolo (approved):
LGTM

PR #1366, opened 2023-07-18
Status for Launcher Connection

mabasian:
To show if the launcher currently has connection, I would propose a status widget of the Launcher <> Server Connection, that is placed right next to the task manager in the footer.


PatrickRL (approved):
LGTM

PR #1369, opened 2023-07-19
FIX: ensure curl for checkpoint availability check

### LINKED ISSUE #1365: [Bug]: invalid checkpoint availability check
### Describe the bug!

Checkpoint availability checks was added in RC20 with #1290. This is done by requesting the CP state via CURL, however if CURL is not installed on the OS the CP check will always fail, thus reporting all CP as invalid and no CP can be selected.

Note that the installer is running before Ansible is executed during installation and therefore CURL is also not auto-installed.

### Attach a Screenshot!

grafik

### Expected Behaviour

Availability is checked and at least one CP is selectable

### Help us reproduce your bug!

- Setup a fresh VM
- Uninstall CURL (if installed): `sudo apt-get remove curl`
- Login to the VM with Stereum and follow oneclick installer until checkpoint selection
- Try to select a checkpoint

### Launcher Version

2.0.0-rc.20

### Desktop OS

All

### Node Server OS

Ubuntu 22.04

### Affected Service

_No response_

### Relevant log output

_No response_

### Acknowledgment

- [X] I looked through the existing bugs of this project

daverolo:
- ensures curl is installed for checkpoint availability check
- if curl is for whatever reason not installable all checkpoints considered as available
- fixes #1365

### DIFF
--- a/launcher/src/backend/Monitoring.js
+++ b/launcher/src/backend/Monitoring.js
@@ -25,6 +25,7 @@ export class Monitoring {
     this.serviceManagerProm = new ServiceManager(this.nodeConnectionProm);
     this.validatorAccountManager = new ValidatorAccountManager(this.nodeConnection, this.serviceManager);
     this.isLoggedIn = false;
+    this.triedCurlInstall = false;
     this.rpcTunnel = {};
     this.wsTunnel = {};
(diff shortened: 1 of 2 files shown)


mabasian (approved):
LGTM!

PR #1370, opened 2023-07-19
additional button to have both "Apply" and "Apply & Restart"

mabasian:
1.  Change the current button name to "**Apply & Restart**"
2. Add another button to the client / service settings, that can be used to run the old function of just simply applying the config, next the current implementation of "Apply & Restart"


PatrickRL (approved):
LGTM

Issue #673, opened 2022-08-17
Complete Dashboard Wiring - (Left to do) Overview


daGscheid:
Please don't close this too prematurely. You can pass things around :) 

- [x] Storage
![image](https://user-images.githubusercontent.com/82385103/185245183-add9236d-768f-4b6f-bb2e-908d496828d6.png)

- [x] Server Connection
![image](https://github.com/stereum-dev/ethereum-node/assets/82385103/caa74e9c-d518-45bf-9600-a070dd9aa95d)

- [x] Network & Node

![NODE PROCESS](https://github.com/stereum-dev/ethereum-node/assets/82385103/fa830b56-0009-41da-8b09-e6f611db7a3a)

One Epoch contains 32 Slots. These Slots are in this component visualized by the square. A slot can be attested (green) or missed (red) by the network participants. This is the chain`s view. 
Every Epoch goes from < PROPOSED > to < JUSTIFIED > to < FINALIZED > 
Slots fill up -> 32 Slots  = 1 Epoch in Progress -> Epoch filled; new Epoch added on top; Epoch is justified for two cycles voting concluded  > Epoch is finalized 

- [x] Blockchain 

![image](https://user-images.githubusercontent.com/82385103/216825473-55a3d097-7d70-481e-93d5-698b9dfb1c50.png)

- [x] Balance Changes
 ![image](https://github.com/stereum-dev/ethereum-node/assets/82385103/72d8bd65-84f2-4b6d-971f-f0b5edf0aea5)

- [x] Validator Stats 
![image](https://github.com/stereum-dev/ethereum-node/assets/82385103/86ede43a-06f4-4854-96e7-fc2bb27a8c7f)

- [x] RPC Endpoint 
![image](https://user-images.githubusercontent.com/82385103/185246194-e4219e8c-64cc-4559-9295-1688188b7ed9.png)

- [x] DATA-API
![image](https://user-images.githubusercontent.com/82385103/185246538-4ef69f0d-1fd9-4691-bae9-f4a0bf90a5d9.png)

- [x] P2P Connections
![image](https://user-images.githubusercontent.com/82385103/185246617-1c08ba3b-2db4-4217-9b07-426303b2182a.png)

daverolo:
Style of "Validator Stats" Widget changed in PR #1000

daGscheid:
As discussed with Stefan - we will introduce some visual changes to the Sync Dashboard which are reflect above, and we will work on with @mabasian  

[...]

PR #1374, opened 2023-07-24
Redesign the Network and Node Component in the control tab

mabasian:
One Epoch contains 32 Slots. These Slots are in this component, visualized by the square. A slot can be attested (green) or missed (red) by the network participants. This is the chain's view.

issue: https://github.com/stereum-dev/ethereum-node/issues/673


gbayasgalan (dismissed):
LGTM 😄

gbayasgalan (dismissed):
nice catch 👍

gbayasgalan (approved):

PR #1379, opened 2023-07-27
FIX: SSV Modal Navigation

### LINKED ISSUE #967: SSV - Network > Navigation Improvements
This is currently a little tedious, because when opening the SSV Naviation Modal, the user keeps getting asked to either provide or copy the Operator Private / Public Key. 

https://docs.ssv.network/learn/introduction

This should only be asked the first time. We could achieve this by asking the SSV Node Operator Registary via its API, if the operator key is already installed. 

If this is the case, the user should be lead to the ssv navigation module

NeoPlays:
## SSV Changes
- Fixes #967
- Updated SSV Grafana Dashboards


daverolo (approved):
LGTM

PR #1383, opened 2023-08-03
ADD: customizable prometheus configuration

PatrickRL:
When applying changes to prometheus configuration, it now adds 'overwrite: false' to prevent automatic overwrite

### DIFF
--- a/launcher/src/components/UI/the-node/TheExpert.vue
+++ b/launcher/src/components/UI/the-node/TheExpert.vue
@@ -59,7 +59,7 @@
           v-for="(option, index) in item.expertOptions.filter((option) => option.type === 'select')"
           :key="index"
           class="selectBox"
-          :class="{ unvisible: isExpertModeActive}"
+          :class="{ unvisible: isExpertModeActive || prometheusExpertModeActive }"
         >
           <img class="titleIcon" :src="option.icon" alt="icon" />
           <span>{{ option.title }}</span>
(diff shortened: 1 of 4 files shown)


launcher/src/components/UI/the-node/TheExpert.vue
```diff
@@ -49,7 +59,7 @@
           v-for="(option, index) in item.expertOptions.filter((option) => option.type === 'select')"
           :key="index"
           class="selectBox"
-          :class="{ unvisible: isExpertModeActive }"
+          :class="{ unvisible: isExpertModeActive && prometheusExpertModeActive }"
```
NeoPlays, line 62: Here you use &&

launcher/src/components/UI/the-node/TheExpert.vue
```diff
@@ -90,7 +100,7 @@
           v-for="(option, index) in item.expertOptions.filter((option) => option.type === 'text')"
           :key="index"
           class="toggleTextBox"
-          :class="{ unvisible: isExpertModeActive }"
+          :class="{ unvisible: isExpertModeActive || prometheusExpertModeActive}"
```
NeoPlays, line 103: but here `||`

NeoPlays (changes requested):
as these conditionals should do the same i think they also should be the same ;)

NeoPlays (approved):
lgtm

Issue #1003, opened 2023-01-09
FR: Login - Security Enhancements 


daGscheid:
Build onto the current server login process by making it possible to manage SSH keys inside the GUI of Stereum by 
- [x] Using SSH keys with password for login 
- [x] Changing SSH keys of the node #
- [x] Change password 

- [x] Creating SSH Key (per @NeoPlays suggestion)

NeoPlays:
Another Idea that came to my mind would be to give the user a way to create SSH keys in the login screen

daGscheid:
Adding a Server Access Management Modal

On first login accessible by clicking here;
![INSTALLATION_MENU – 16](https://github.com/stereum-dev/ethereum-node/assets/82385103/e6667d25-121e-418a-a05e-28f0c6398f89)



After node installation accessible by clicking here 
![Screenshot 2023-08-16 111705](https://github.com/stereum-dev/ethereum-node/assets/82385103/df997917-1b31-4c26-81d3-3910e880cf8c)


Server Access Management Modal

Server utilizing Password Access (No SSH Key)
![Node Page – 20](https://github.com/stereum-dev/ethereum-node/assets/82385103/a76b04b0-9f74-4ba0-ad61-9e0562efae4d)

Server utilizing SSH Key 
![Node Page – 25](https://github.com/stereum-dev/ethereum-node/assets/82385103/2d108359-ce33-4685-966a-ebdf7fa91975)

Remove the SSH Key from the server
![Node Page – 31](https://github.com/stereum-dev/ethereum-node/assets/82385103/692a1311-203f-4020-aa2b-4fbe824f1e10)

Generate a SSH Key
![Node Page – 29](https://github.com/stereum-dev/ethereum-node/assets/82385103/46b7c052-8adb-4bae-9ab9-b72da668a7c1)
![Node Page – 30](https://github.com/stereum-dev/ethereum-node/assets/82385103/895d0a4b-2ef9-4a2a-a242-c66fe6d04c58)

daGscheid:
Available with rc.22, good job guys

PR #1384, opened 2023-08-04
ADD: SSH Features

NeoPlays:
First Step of #1003

Also fixed data retention getting overwritten with update.

So this PR adds functions for:
- changing password
- generating SSH Keys
- reading SSH Keys
- writing SSH Keys


daverolo (approved):
LGTM

Issue #1353, opened 2023-07-13
[FR]: Include the option to "ignore" Doppelganger Protection on Validator Client Import (for Expert Users)


daGscheid:
### Describe your idea

Make it an option in the validator client settings to deactivate "default doppelganger protection".
If then deactivated, the import will throw an additional prompt asking if one wants to use doppelganger protection on import.

- [x] Discuss
- [x] Design
- [x] Build

### Considered Alternatives?

--

### Additional Context

--

### Acknowledgment

- [X] I looked through the feature requests/enhancements planned for this project before submitting a new idea

PR #1387, opened 2023-08-05
Doppelganger protection

gbayasgalan:
issue: https://github.com/stereum-dev/ethereum-node/issues/1353

- doppelganger protection added into all validator clients as default. 
- DPs could be disabled/enabled using Expert Mode

many thanks @daverolo for your time:) 🚀


launcher/src/store/services.js
```diff
@@ -170,6 +177,13 @@ export const useServices = defineStore("services", {
               icon: "/img/icon/plugin-menu-icons/fee.png",
               pattern: "(--suggested-fee-recipient=)(.*)(\\n)",
             },
+            {
+              title: "Doppelganger",
+              type: "toggle",
+              changeValue: true,
+              icon: "/img/icon/plugin-menu-icons/doppelganger.png",
+              pattern: "(--enable-doppelganger=)(.*)(\\n)",
```
daverolo, line 185: seems you have used the old pattern here. correct pattern would be the same as on line 97. this counts also for patterns on line 282, 387 and 821

daverolo (changes requested):
please check patters in service.js @gbayasgalan

launcher/src/store/services.js
daverolo: ok, those are the entries from the other clients, so please forget it. sunday morning is probably the wrong time for @daverolo to do a review ;)

daverolo (approved):
LGTM 🚀

PR #1389, opened 2023-08-07
Update Amsterdam comp. with the Gnosis Network

mabasian:
When operating on an Ethereum network following the ETH consensus, the component aligns with an epoch of 32 slots, each updating every 12 seconds. In contrast, when interacting with the Gnosis network, the component adjusts to an epoch length of 16 slots and a faster slot update interval of 6 seconds.


daverolo (approved):
LGTM 🚀

PR #1398, opened 2023-08-09
Fix the call slot function because call twice it

mabasian:


gbayasgalan (approved):
nice thank you @mabasian 🚀

PR #1405, opened 2023-08-11
previous justified epoch status

gbayasgalan:
previous justified epoch status
issue: https://github.com/stereum-dev/ethereum-node/issues/673


mabasian (approved):
LGTM🚀

PR #1408, opened 2023-08-16
redo the Amsterdam comp.

mabasian:
Redo the Amsterdam component to add pre justify row and also create a dialog on top of the widget to be more user-friendly


gbayasgalan (approved):
lgtm 👍

PR #1409, opened 2023-08-16
ADD: format check, lockfile delete button

### LINKED ISSUE #1050: FR: Warning & path to resolve Lockfile Creation issues
**Is your feature request related to a problem? Please describe.**
User can run into a problem with lockfiles being already in use (through unexpected shutdown f.e.) 

**Describe the solution you'd like**
A warning if there are problems & way to disable the creation of the lockfiles by setting this flag, and resolve the problem / or alternatively a link to documentation that describes the process to do that. 

Example: If Teku dies the container doesn't automatically start anew through this problem

### LINKED ISSUE #1272: [FR]: QoL Improvement - Prevent that a wrong format in Expert Mode breaks the Node
### Describe your idea

While the "Expert Mode" kind of asumes a certain knowledge, mistakes happen to us constantly as well. Now that bystro, a community member ran into the same problem, we had a shared thought in the team and came to - quite obvious one has to say - conclusion, that it is really not necessary that a wrong format breaks the node completly. 

For this we could check the format of the configs & notify the user of this. 

### Considered Alternatives?

Leave as is - it is a good QoL, otherwise it gets annoying fast. 

### Additional Context

image

They should know & it really shouldn't necessary. 

### Acknowledgment

NeoPlays:
## Option to remove Teku Lockfiles
The Teku Valdiator Export Mode now includes a button which removes validator lockfiles. this should make it easier for users to resolve their teku issues after teku crashes. fixes #1050

## Check for wrong service formats
Changes to service configs via Expert Mode are now checked for the correct format before writing.
Tht `write yaml` task will fail if the format is not correct. fixes #1272

### DIFF
--- a/launcher/src/backend/NodeConnection.js
+++ b/launcher/src/backend/NodeConnection.js
@@ -568,6 +568,9 @@ export class NodeConnection {
     const ref = StringUtils.createRandomString();
     this.taskManager.tasks.push({ name: "write yaml", otherRunRef: ref });
     try {
+      if (!this.checkServiceYamlFormat(service.data.trim())) {
+        throw new Error("Config is not in the right format!");
+      }
       configStatus = await this.sshService.exec(
         "echo -e " +
         StringUtils.escapeStringForShell(service.data.trim()) +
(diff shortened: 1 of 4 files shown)


daverolo (approved):
LGTM

PR #1417, opened 2023-08-22
REFACTOR: os check display

### LINKED ISSUE #1364: [FR]: OS Check Display Changes
### Describe your idea

Change the display of the OS Check after logging in & when first choosing between installation options to

- [ ] instead of blocking the user when the are installing anything but Ubuntu 22.4, we should let them install version 23.04 f.e. with a warning, that says we recommending a certain version number
- [ ] let the user user install Stereum on Non Ubuntu OS, instead of completly blocking them. 

### Considered Alternatives?

--

### Additional Context

--

### Acknowledgment

daverolo:
- remove centos as supported os
- implement exact version check for supported os (currently ubuntu 22.04)
- reduce loading time on os check
- show warning but allow installation even on unsupported os
- fixes #1364

### DIFF
--- a/launcher/src/backend/NodeConnection.js
+++ b/launcher/src/backend/NodeConnection.js
@@ -20,6 +20,7 @@ export class NodeConnection {
     this.sshService = new SSHService();
     this.nodeConnectionParams = nodeConnectionParams;
     this.os = null;
+    this.osv = null;
   }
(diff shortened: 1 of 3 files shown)


NeoPlays (approved):
lgtm 🚀

PR #1421, opened 2023-08-24
ADD: Archive node option

PatrickRL:
Added archive node option in One Click Installation

Changed Reth Execution Client to only show in the new archive node option, since it is archive only.


NeoPlays (approved):
good job 🚀

Issue #1220, opened 2023-05-10
[Bug]: Add a Network Selection to Custom Installation


daGscheid:
### Describe the bug!

When deleting the node & trying to install another one with Custom Installation it takes into  account the last network one had selected, when using it. This can lead to really weird behaviour. 

We can fix this, by including a Network Selection in the Custom Installation Modal. 

- [ ] Include Network Selection Dropdown Box in Custom Installation Modal

![INSTALLATION_MENU – 3](https://github.com/stereum-dev/ethereum-node/assets/82385103/c5a1ef1c-34dd-4311-a494-da7a9aca8d7e)

### Attach a Screenshot!

CURRENT VERSION: 
![image](https://github.com/stereum-dev/ethereum-node/assets/82385103/0995c948-b90b-4c36-a84f-24b026543c3d)

### Expected Behaviour

I expected it to be possible to change the network! 👯 

### Help us reproduce your bug!

1) Install any Node with One Click Installation
2) Nuke the Node
3) Use Custom Installation 
4) The last network selected in One Click Installation is used as the network selected. 

F.e. This can lead the user not to be able to select Gnosis when they had selected an Ethereum Testnet before (& vice versa

### Launcher Version

2.0.0-rc.17

### Desktop OS

Windows

### Node Server OS

Ubuntu 22.04

### Affected Service

-----

### Relevant log output

```shell
----
```

### Acknowledgment

- [X] I looked through the existing bugs of this project

PR #1422, opened 2023-08-25
Add a Network Selection to Custom Installation

mabasian:


mabasian:
I fixed #1220 on the frontend, but I need help from a backend developer to fully fix it. I tried to do the same networkSwitch component, but it has a different state and I don't know exactly the JSON structure that needs to be sent to the server to set the change. Likewise, I can do it, but it takes time, with the help of a backend developer, it can be fixed faster, and that's why I assigned all the backend developers, but I think it needs only the help of one of them. 😅

daverolo (approved):
LGTM

PR #1432, opened 2023-08-31
REMOVE: Bloxroute ethical relay

NeoPlays:
- Bloxroute ethical relay removed
- Restart failed sometimes on setup
- Monitoring loop tried to get service configs and failed due to missing `/etc/stereum/services` directory. On OneClickInstallations Stereum now creates not only `/etc/stereum` but `/etc/stereum/services` as well


daverolo (approved):
LGTM

PR #1434, opened 2023-08-31
FIX: restart/resync of services

PatrickRL:
Fixed a bug were services which haven't been started once would throw an error on restart or resync


NeoPlays (approved):
lgtm 🚀

PR #1435, opened 2023-09-01
Update the internationalization and footer tooltip

mabasian:
New modal text was added to language JSON to translate, and the tooltip was also updated. The Czech language enabled.


daverolo (approved):
LGTM

PR #1394, opened 2023-08-08
improve functionality of switch button on the settings

mabasian:
The buttons know what the current state of the corresponding settings is, and react with the correct state of the settings, not just once the settings page is loaded.
Fix issue #1392

PR #1471, opened 2023-09-26
FIX: Pruning and Subtasks

NeoPlays:
This PR reverts #1453 because it causes Subtasks Details do close instantly on opening them.
This PR also reverts #1394 because it broke Geth Pruning.


mabasian (approved):
LGTM

PR #1485, opened 2023-10-06
ADD Option to NOT automatically Start the Client after OneClickInstallation 

mabasian:


MaxTheGeeek:
@daGscheid @mabasian  what is this? we do not have plugin name component anymore.

PR #1494, opened 2023-10-19
editing external ip and port, also doppelganger protection for new UI

gbayasgalan:
on the expert mode needs some UI improvement @mabasian (same as what we did on old one)


MaxTheGeeek (approved):
lgtm

PR #1521, opened 2023-11-02
FIX: structured clone issue

NeoPlays:
the function `structuredClone` is not able to clone refs

So i added the function `useDeepClone`.


MaxTheGeeek (approved):
lg

PR #1528, opened 2023-11-03
Resolve Issue with Client Buttons

mabasian:
After adding tooltips, the buttons had issues and didn't work initially, but the problem has been resolved and they are functioning correctly now


gbayasgalan (approved):
lgtm

PR #1593, opened 2023-12-11
ADD: Dump Docker Logs Functionality

NeoPlays:
Connected the backend function with the frontend.
also improved the way network stats where handled for checkpoint sync


mabasian (approved):
LGTM

PR #1594, opened 2023-12-11
UPDATE: Erigon Config

NeoPlays:
Added Erigon Tags `prune=htc`,`prune.r.before`,`--rpc.returndata.limit=1000000`, `--db.pagesize=16K'` and `--db.size.limit=8TB`

Also added Update Script for tag `db.size.limit=8TB`

Erigon now available for gnosis

Erigon Archive Config fixed


daverolo (approved):
LGMT 🚀

Issue #1095, opened 2023-02-23
BUG: Prometheus NE shows wrong value for used storage


daverolo:
**Describe the bug**
Prometheus NE shows wrong value for used storage in dashboard wIdget

**To Reproduce**
Setup a fresh node and check the Storage Widget

**Expected behavior**
Corret value for Prometheus NE used storage

**Screenshots**
![grafik](https://user-images.githubusercontent.com/107847185/221043335-b6ceefb0-90c3-47ff-8895-4e26f70b1efc.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Launcher version: 2.0.0-rc.13

**Node (please complete the following information):**
 - OS: Ubuntu 22.04
 - Affected service: Prometheus NE

**Additional context**
this value is calculated from the docker volumes and the node exporter needs to map the root directory so it very likely takes the size of the root dir. prometheus node exporter stores no data.

daverolo:
Since Prometheus NE maps the root ("/") directory it automatically shows the size of it. However, in fact it does not store any data and uses 0 byte. Instead of showing it with 0 B like this ...
![grafik](https://user-images.githubusercontent.com/107847185/225600283-c2178464-901f-4d8d-89c7-c413f6cca54b.png)
**... we decided to entirely remove it from the storage list.**

daverolo:
Finally entirely removed from the storage query with #1602 for performance reasons.
Does not make sense to run `du -csh "/" | tail -n1 | awk '{print $1;}'` which just spams the logs and consumes resources while NE is already disabled in the storage list for the front-end.

PR #1602, opened 2023-12-20
REFACTOR: remove nodexporter from storage query

daverolo:
Remove Prometheus NodeExporter service from storage query (extends #1095)


gbayasgalan (approved):
Lgtm 👍

Issue #1451, opened 2023-09-13
[Bug]: Slot viewer incorrect data


stefa2k:
### Describe the bug!

The slot view displays the current epoch and the previous epoch (justified) correctly. But the previous justified epoch and the finalized epoch is the same, see screenshot.

### Attach a Screenshot!

![image](https://github.com/stereum-dev/ethereum-node/assets/54934211/a2c204e0-1d4e-4833-b93d-b501a81bf12b)

### Expected Behaviour

Having the justified and finalized differently

### Help us reproduce your bug!

Nimbus, Nethermind (didn't try other clients)

### Launcher Version

2.0.0-rc.22

### Desktop OS

Mac

### Node Server OS

Ubuntu 22.04

### Affected Service

Launcher

### Acknowledgment

- [X] I looked through the existing bugs of this project

gbayasgalan:
Actually they are written separately in the code. 
The `Beacon-API` is used to catch `Epochs` & The `Beacon-API` obtain the same numbers for both `previous justified` & `finalized` epochs. I guess that's how beacon-API works so far (at least in my experience I always got same numbers for them)

Try following command to have a look on beacon-API output: 
`curl -s -X 'GET' 'http://localhost:${beaconAPIPort}/eth/v1/beacon/states/head/finality_checkpoints' -H 'accept: application/json'`

e. g: 
`{
  "execution_optimistic": true,
  "finalized": false,
  "data": {
    "previous_justified": {
      "epoch": "203435",
      "root": "0x37938e0c23a168bab490bc316a312b4c5aa06fa627391ec582c204bc1a298c66"
    },
    "current_justified": {
      "epoch": "203436",
      "root": "0x6af3c7670e316481bd141cebe3a3d58b28e5acc994c7c84bbec12fad176a773d"
    },
    "finalized": {
      "epoch": "203435",
      "root": "0x37938e0c23a168bab490bc316a312b4c5aa06fa627391ec582c204bc1a298c66"
    }
  }
}`

gbayasgalan:
Beacon-API retrieved finalized Epoch Number is manually reduced once (tmp solution :) )

PR #1605, opened 2023-12-21
FIX: distinguish finalized & preJustified epochs on Control tab

gbayasgalan:
fix: https://github.com/stereum-dev/ethereum-node/issues/1451
finalized Epoch from beacon-API is just manually reduced once


mabasian (approved):
LGTM

PR #1609, opened 2023-12-27
UPDATE: electron.yaml

NeoPlays:
As It seams do you need to update `actions/download-artifact` together with `actions/upload-artifact`
And as `actions/upload-artifact` is fixed with `v2` and `actions/download-artifact` is always at `master` the versions mismach. https://github.com/actions/download-artifact/issues/267


daverolo (approved):
LGTM

PR #1734, opened 2024-03-02
FIX: os updates in updatepanel

daverolo:
- Improve loading speed
- Fix style of OS updates badge
- Fix variable: numberOfUpdatablePackages
- Fix method: getUpdatablePackagesCount()
- Remove method: getUpgradablePackages()

### DIFF
--- a/launcher/src/components/UI/base-header/components/modals/UpdatePanel.vue
+++ b/launcher/src/components/UI/base-header/components/modals/UpdatePanel.vue
@@ -41,9 +41,9 @@
                       />
                       <div
                         v-else
-                        class="w-12 h-[17px] bg-red-700 rounded-full p-1 text-[12px] text-gray-200 text-center flex justify-center items-center mr-2"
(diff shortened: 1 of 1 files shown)


launcher/src/components/UI/base-header/components/modals/UpdatePanel.vue
```diff
@@ -339,37 +339,23 @@ const searchOsUpdates = async () => {
   searchingForOsUpdates.value = true;
   setTimeout(async () => {
```
NeoPlays, line 340: is this `setTimeout` even needed?

NeoPlays: i guess if the function is "to fast" the user will have no response from the UI but 10 ms will be to fast as well ^^

daverolo: You're absolutely right, it's not needed but since it did exist, i just kept it there with a minimum timeout to let others decide what to do with it :)

NeoPlays (approved):
lgtm

Issue #1717, opened 2024-02-26
[FR]: Shell Access in Stereum UI (QoL) with Predefined Commands


daGscheid:
### Describe your idea

Original idea by my man @daverolo, who dropped it today. During testing for SSV & Obol, he constantly had to switch between Stereum and the Shell. This is very annoying as he mentioned and one might imagine. 

Now why don't we just include a shell page with predefined commands, making the constant switching unnecessary and open even more freedom to the expert user inside of Stereum.
(Note: Might make it necessary to hide from beginners, we can discuss this)

##### How to access: 

![image](https://github.com/stereum-dev/ethereum-node/assets/82385103/515c3dfc-1f5a-4598-9178-e871c7af6083)

For easy access I would suggest to put this into the header

##### The Secure Shell (SSH) Page:

![Node Page – 17](https://github.com/stereum-dev/ethereum-node/assets/82385103/4b41f48b-0942-4472-8af9-1703471892bb)

### Considered Alternatives?

--

### Additional Context

--

### Acknowledgment

- [X] I looked through the feature requests/enhancements planned for this project before submitting a new idea

PR #1737, opened 2024-03-04
ADD: Shell Access

gbayasgalan:
Fixes https://github.com/stereum-dev/ethereum-node/issues/1717


gbayasgalan:
jest test is failing for now because of new ansible role which is not merged into the main branch, but is called in jest test from backend. hopefully merge could happen soon. atm tested with local branch and it works fine :)

NeoPlays (approved):
very nice! 🚀

PR #1775, opened 2024-03-18
FIX: running image version got lost with new design

NeoPlays:
the image version shown in the header of the logs, was once the actual running image version. this probably got lost in the transition to the new design, so i fixed this in this pr.


mabasian (approved):
LGTM

PR #1776, opened 2024-03-18
Add: Molecule Tests

M4r71nW:


controls/roles/switch_repos/molecule/default/molecule.yml
```diff
@@ -0,0 +1,27 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: molecule_hetznercloud
+platforms:
+  - name: "update-os--default--ubuntu-22.04"
```
NeoPlays, line 7: The instance name should always have the following format:
`<role-name>--<scenario>--<distro + version>`

could you pls update this accordingly? 🙏🏼

NeoPlays (changes requested):
lgtm 🚀 , pls just update the molecule.ymls

NeoPlays (approved):
🚀

PR #1780, opened 2024-03-19
ADD: LidoObolExitService

NeoPlays:
Adds LidoObolExitService which automatically creates signed messages for the ValidatorEjectorService.


launcher/src/backend/ethereum-services/LidoObolExitService.js
```diff
@@ -0,0 +1,57 @@
...
+
+    service.init(
+      "LidoObolExitService", //service
+      service.id, // id
+      1, // configVersion
+      image, // image
+      "latest", // imageVersion
+      [
+        "run",
+        "--beacon-node-url=http://lighthouse:5052",
```
gbayasgalan, line 28: only Lighthouse?

NeoPlays: this is just a placeholder for the beacon service

gbayasgalan: since its just placeholder. imho, to use something like `http://beaconNode:beaconPort` is better, isn't it?

NeoPlays: sure, but as i'm the only one using it is just went for the quick and dirty sollution

gbayasgalan: got it 😄

gbayasgalan (approved):
lgtm 👍

Issue #1774, opened 2024-03-17
[Bug]: Update Arrow in Header always green despite no Update available


daGscheid:
### Contact Details

--

### Describe the bug!

The green "up-arrow" button in the header of Stereum is always green even if there's no new update available

### Attach a Screenshot!

https://github.com/stereum-dev/ethereum-node/assets/82385103/863ee596-7bfb-4ba2-b910-64d9b1530db7

### Expected Behaviour

I expected the green up-arrow icon only to show when there is an update available. 
Otherwise it should be blue. 

### Help us reproduce your bug!

Here is how to reproduce it!

### Launcher Version

2.1.0

### Desktop OS

All

### Node Server OS

Ubuntu 22.04

### Affected Service

--

### Relevant log output

```shell
--
```

### Acknowledgment

- [X] I looked through the existing bugs of this project

daGscheid:
Fixed via: [Fix the header update icon and Fee Recipient Option in settings #1781](https://github.com/stereum-dev/ethereum-node/pull/1781)

Issue #1773, opened 2024-03-17
[Bug]: Fee Recipient Option in Validator Client settings broken


daGscheid:
### Contact Details

daGscheid

### Describe the bug!

Since the 2.1.0 update the defailt fee recipient is no longer editable via in the validator client settings 

### Attach a Screenshot!

https://github.com/stereum-dev/ethereum-node/assets/82385103/5184bc04-70fb-46f4-92ce-2175ee80e2ea

### Expected Behaviour

--

### Help us reproduce your bug!

1) Acess Node Page
2) Open Validator Client Settings

### Launcher Version

2.1.0

### Desktop OS

All

### Node Server OS

Ubuntu 22.04

### Affected Service

--

### Relevant log output

```shell
--
```

### Acknowledgment

- [X] I looked through the existing bugs of this project

daGscheid:
Fixed via: [Fix the header update icon and Fee Recipient Option in settings #1781](https://github.com/stereum-dev/ethereum-node/pull/1781)

PR #1781, opened 2024-03-19
Fix the header update icon and Fee Recipient Option in settings

mabasian:
Fix issue #1774 
Fix issue #1773


daGscheid (approved):
Works, thank you

PR #1849, opened 2024-04-25
FIX: Login Issues

NeoPlays:
Logouts were processed asynchronously and new connections were tried to create without a finished logout, resulting in weird issues.


daverolo (approved):
🚀

PR #1857, opened 2024-05-02
ADD: OS Upgrade

M4r71nW:


launcher/src/components/UI/server-management/components/server-update/UpdateServer.vue
```diff
@@ -98,15 +108,37 @@ const stereumApp = ref({
 const numberOfUpdatablePackages = ref(null);
 const searchingForUpdatablePackages = ref(false);
 
+console.log("getAllTasks", serverStore.allTasks);
```
NeoPlays, line 111: i think we can remove that.

launcher/src/components/UI/server-management/components/server-update/UpdateServer.vue
```diff
@@ -98,15 +108,37 @@ const stereumApp = ref({
...
 const newUpdates = computed(() => serverStore.upgradablePackages);
 const onOff = computed(() => (stereumApp.value.autoUpdate == "on" ? "text-green-700" : "text-red-700"));
 
+//Watchers
+
+watchEffect(() => {
```
NeoPlays, line 117: i think `watchEffect` is a bit overkill here. you can also use `computed`

launcher/src/backend/NodeUpdates.js
```diff
@@ -19,7 +19,7 @@ export class NodeUpdates {
       await this.nodeConnection.runPlaybook("upgrade-prep", { stereum_role: "upgrade-prep" });
       log.info("Starting Upgrade to noble numbat ...");
       const result = await this.nodeConnection.sshService.exec(
-        "do-release-upgrade -d -f DistUpgradeViewNonInteractive --allow-third-party"
+        "do-release-upgrade -f DistUpgradeViewNonInteractive --allow-third-party && shutdown now -r"
       );
       await this.nodeConnection.restartServer();
```
NeoPlays, line 24: you'll need to adapt here:

- You can `touch /var/run/reboot-required` insead of `shutdown now -r`
- You can adapt the restartServer method so , that you pass a tag something like `forceReboot` so it skips the check for reboot required and runs the reboot anyway.

NeoPlays (changes requested):
some smaller stuff to check, but looks good otherwise 🚀

launcher/src/components/UI/server-management/components/server-update/UpdateServer.vue
NeoPlays: So from the [docs](https://vuejs.org/guide/essentials/watchers.html#watcheffect) it clearly shows that watchers are used for refs that change on a regular basis, so it can adapt on chages immediately. In this case `osVersionCurrent` will change twice a year at max. Users might not even upgrade at all.

PR #1864, opened 2024-05-07
FIX: Validator Import with Password Files

NeoPlays:
If importing more than 100 keys with password files, only the first 100 would be imported. A wrong password batch was assigned for the other batches


daverolo (approved):
LGTM

PR #1867, opened 2024-05-11
FIX: ssvmodal support for SSV 1.3.3 keystore format

daverolo:
This PR will support the keystore format that has changed in SSV version 1.3.3


mabasian (approved):
LGTM

PR #1904, opened 2024-06-04
Fix Naming of Doppelganger to Doppelganger Protection in settings

daGscheid:
Change title of option in validator client settings from "Doppelganger" to "Doppelganger Protection"

### DIFF
--- a/launcher/src/store/services.js
+++ b/launcher/src/store/services.js
@@ -117,7 +117,7 @@ export const useServices = defineStore("services", {
               commands: ["--suggested-fee-recipient"],
             },
             {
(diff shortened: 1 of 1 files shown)


M4r71nW (approved):

launcher/src/store/services.js
```diff
@@ -117,7 +117,7 @@ export const useServices = defineStore("services", {
               commands: ["--suggested-fee-recipient"],
             },
             {
-              title: "Doppelganger",
+              title: "Doppelganger Protection",
```
gbayasgalan, line 120: only for LH?

PR #1905, opened 2024-06-05
FIX: Naming of Doppelganger to Doppelganger Protection in settings for rest of the Clients

gbayasgalan:
Naming of Doppelganger to Doppelganger Protection in settings for rest of the Clients


daGscheid (approved):
LGTM

PR #1938, opened 2024-07-09
Improve No Data Visualization

### LINKED ISSUE #1654: [FR]:  Improve No Data Visualization
### Describe your idea

- [x] Better visualize when there is no Consensus Client OR Validator Client installed. It looks like the user can expect data that will never arrive. We should add the missing "NoData" States @mabasian
- [x] Move the text "Please turn on your Consensus Client, Prometheus and Grafana" into the tooltips

image

### Considered Alternatives?

--

### Additional Context

--

### Acknowledgment

mabasian:
FIX #1654


MaxTheGeeek (approved):
lg

PR #1940, opened 2024-07-12
FIX: ssv keychange

daverolo:
Reset existing SSV database if operator key changes on:
- importing an encrypted SSV key
- importing an unencrypted SSV key
- importing an (old) unencrypted SSV key and migrate this key to a (new) encrypted SSV key
- creating a new encrypted SSV key


mabasian (approved):
lgtm

PR #1943, opened 2024-07-16
FIX: couple issues on Staking Page

NeoPlays:
- fixes some smaller stuff on Staking page (still not perfect)
- removed loading when entering Control / Node Page
- fixed small bug on edit page


launcher/src/components/UI/base-header/components/modals/NotifModal.vue
```diff
@@ -165,7 +165,7 @@ export default {
       let test = [];
       const selectedSetup = this.selectedSetup;
       if (selectedSetup && selectedSetup.services) {
-        const selectedServiceIds = selectedSetup.services.map((service) => service.id);
+        const selectedServiceIds = selectedSetup.services.map((service) => service.config.serviceID);
```
MaxTheGeeek, line 168: is it correct? we do NOT have service.config.serviceID in the selectedSetup.

NeoPlays: yes we do, because i changed it. the services in selectedSetup is now the same data as in InstalledServices.

before we had to manage 2 different service states which not only broke stuff, but also made everything very hard.

mabasian (approved):
LG

PR #1987, opened 2024-08-12
FIX: missing ssv.network service on holesky edit node

daverolo:
Fixes the missing ssv.network service  in "_Edit Node -> Add Service_" dialog on Holesky.


NeoPlays (approved):
lgtm

PR #2037, opened 2024-09-04
FIX: OS Updates and Alerts

### LINKED ISSUE #2019: [Bug]: OS Updates and Alerts
### Environment

Development

### Bug Description

Available OS Updates are not refreshed after updating.

An Alert Icon is not loading
grafik

### Launcher Version

2.2.2

### Operating System

Mac

### Acknowledgment

- [X] I have checked the existing bugs and confirmed this issue is not already reported.

### LINKED ISSUE #2034: [Bug]: OS Updates list does not refresh
### Environment

Development

### Bug Description

1. Executed OS updates
image
2. After the update was done the machine rebooted and Stereum automatically re-connected to the node.
However, the update count and the update list was still the same as before the update/reboot.
image
image
3. Logged out manually
image
4. On next login the update list was as expected
image
image

### Launcher Version

main:39b902327ce03a2ccb90fb604279240c842a5e5f

### Operating System

Mac

### Affected Service (if applicable)

OS updates (panels)

### Logs (if applicable)

_No response_

### Acknowledgment

- [X] I have checked the existing bugs and confirmed this issue is not already reported.

mabasian:
fixes #2019
fixes #2034


NeoPlays (changes requested):
That will not quite work as updating a single pkg may result in other pkges updated.
Also when you update all pkgs in some cases it will still show some pkgs as they're not updateable at that point.

I suggest simply fetching the stats again after updating. i think that's the easiest solution.

PR #2102, opened 2024-10-31
ADD: Idle timer

PatrickRL:
Added option to set an idle timer that logs out the user after a set amount of idle time


NeoPlays (approved):
lgtm

PR #2104, opened 2024-11-07
queued in CSM status check

M4r71nW:
queued in CSM status check:

implementation of "csm-status-draft.pdf"
```
// Example usage

const { checkSigningKeys } = require('./csmodule');

(async () => {
  const nodeOperatorId = 43; // ID you want to check
  const keysArray = ['0xa86f4444b19220782de8127af0cf2be0b2ab58a441956224b52058090198e2a97c0b3548a9d0722a1c5cf46e7c275163',
  '0x8efdbef6c5faf5e7dc5bfcc39979d355e5c86ca6230533c959450ba8285172b1f5a64a03fdc1745ea86efb5f9491b930',
  '0xaab211983b5ac07ba73daa26aac14084af6a0e5483370aee95b1402af0615b38b47c90a7366ff6337290d0c25264e15e']; // keys you want to check

  const result = await checkSigningKeys(nodeOperatorId, keysArray);
  if (result === false) {
    console.log("No matches found for the provided keys.");
  } else if (result === null) {
    console.log("An error occurred while checking keys.");
  } else {
    console.log("Matching keys:", result);
  }
})();

/* Example 1 Output:
Loaded csm_abi successfully
Connected to Mainnet. ID = 1
Node is in sync ... 
Node Operator is active: true
Matching keys: [
  '0xa86f4444b19220782de8127af0cf2be0b2ab58a441956224b52058090198e2a97c0b3548a9d0722a1c5cf46e7c275163',
  '0x8efdbef6c5faf5e7dc5bfcc39979d355e5c86ca6230533c959450ba8285172b1f5a64a03fdc1745ea86efb5f9491b930',
  '0xaab211983b5ac07ba73daa26aac14084af6a0e5483370aee95b1402af0615b38b47c90a7366ff6337290d0c25264e15e'
]

Example 2 Output:
Loaded csm_abi successfully
Connected to Holesky. ID = 17000
Node is in sync ... 
Node Operator is active: true
No enqueued validators for this Node Operator.
No matches found for the provided keys.
*/
```


NeoPlays:
i think we should be able to implement this in [ControlService](https://github.com/stereum-dev/ethereum-node/blob/main/launcher/src/store/ControlService.js) as your functions need to run in the renderer process to have access to the window object. for this we may want to move the ControlService to a separate folder named renderer or something like this so we keep this structured

NeoPlays:
With `await ControlService.getCSMQueue(<ArrayOfKeysToCheck>)` you can check for the queue status.

it will return:
- an `Array` with keys in queue (pubkeys)
- `false` if no passed key is in queue
- `null` if there was an error

MaxTheGeeek (approved):
lgtm

PR #2126, opened 2024-11-27
FIX: Exits for Obol DVT Nodes

NeoPlays:


launcher/src/backend/Monitoring.js
```diff
@@ -3384,8 +3384,12 @@ export class Monitoring {
     }
 
     try {
-      const beaconAPIPort = beaconStatus.data[0].beacon.servicePort;
-      const serviceId = beaconStatus.data[0].sid;
+      //Check if connected to CharonService
+      const vc = await this.nodeConnection.readServiceConfiguration(serviceID);
+      const connectedCharon = vc?.dependencies.consensusClients.find((client) => client.service === "CharonService");
+
+      const beaconAPIPort = connectedCharon ? 3600 : beaconStatus.data[0].beacon.servicePort;
```
gbayasgalan, line 3391: 3600 isn't changeable?

NeoPlays: the port is not exposed, we could get the port via the command, but if you look at everything else in monitoring we expect a default port configured

gbayasgalan (approved):
🚀

PR #2127, opened 2024-12-02
ADD: Stereum Log Backup

PatrickRL:
Automatically create backups of the Stereum log file

Added option to configure how many backups are stored


NeoPlays:
I think you missunderstood the issue. I think the logs you should backup are the service logs not the stereum-launcher logs.
@daGscheid correct me if i'm wrong

NeoPlays (changes requested):
pls use [electron-log file transport](https://github.com/megahertz/electron-log/blob/HEAD/docs/transports/file.md#archivelogfn-oldlogfile-logfile--void ) to handle old log files

launcher/src/components/base/BaseLayout.vue
```diff
@@ -190,6 +190,11 @@ const checkSettings = async () => {
         await ControlService.idleTimerCheck(false);
       }
     }
+    if (typeof savedConfig.logBackups.value !== "undefined") {
```
NeoPlays, line 193: It does not really make sense to store this detail on the server when it's affecting files on the local machine so independent of any server

launcher/src/background.js
```diff
@@ -33,10 +34,24 @@ const tekuGasLimitConfig = new TekuGasLimitConfig(nodeConnection);
...
 log.transports.console.level = process.env.LOG_LEVEL || "info";
 log.transports.file.level = "debug";
+log.transports.file.archiveLogFn = (file) => {
+  file = file.toString();
+  const info = path.parse(file);
+  let backupPath = info.dir + "/backups/";
+  if (!existsSync(backupPath)) {
+    mkdirSync(backupPath);
+  }
+
```
NeoPlays, line 49: the main.log is not renamed or moved, this will not store new logs after 1 mb if i'm not mistaken.

launcher/src/background.js
```diff
@@ -33,10 +34,24 @@ const tekuGasLimitConfig = new TekuGasLimitConfig(nodeConnection);
...
+  copyFileSync(file, `${backupPath}main-${Date.now()}.log`);
```
NeoPlays, line 50: i would also move the delete logic here

NeoPlays (changes requested):

launcher/src/background.js
NeoPlays, line 50: You can just use rename just like in the example i send you then you don't need to copy and then use unlink. (if you're deleting files i'd always go with rm and not unlink)

NeoPlays (approved):
very nice 🚀

Issue #2106, opened 2024-11-13
[Bug]: Removing Setups doesn't delete Recources


NeoPlays:
### Environment

Development

### Bug Description

Container, Data and Configs are not deleted

### Launcher Version

latest

### Operating System

Mac

### Acknowledgment

- [X] I have checked the existing bugs and confirmed this issue is not already reported.

gbayasgalan:
@NeoPlays, seems like it works for me, Could you provide more detailed steps to reproduce the issue?

PR #2159, opened 2024-12-27
FIX: remove setup services right after creating setup

MaxTheGeeek:
fixes: https://github.com/stereum-dev/ethereum-node/issues/2106


gbayasgalan (approved):
lg 😃

PR #2165, opened 2025-01-14
FIX: Escaping Problem of Passwords on Import

### LINKED ISSUE #2163: [Bug]: Password with special Chars fail when trying to import key
### Environment

Release

### Bug Description

- create keystore with a password that includes "`" for example.
- try to import it on staking page
- error

NeoPlays:
fixes #2163

### DIFF
--- a/launcher/src/backend/ValidatorAccountManager.js
+++ b/launcher/src/backend/ValidatorAccountManager.js
@@ -36,8 +36,10 @@ export class ValidatorAccountManager {
     //escape all passwords for shell
     passwords = passwords.map((p) => {
       let pass = p;
-      if (pass.includes('"')) pass = pass.replaceAll(/"/g, '\\"');
-      if (pass.includes("'")) pass = pass.replaceAll(/'/g, `'\\''`);
+      if (pass.includes("\\")) pass = pass.replaceAll(/\\/g, "\\\\");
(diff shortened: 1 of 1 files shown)


launcher/src/backend/ValidatorAccountManager.js
```diff
@@ -33,7 +33,13 @@ export class ValidatorAccountManager {
       }
       return readFileSync(file.path, { encoding: "utf8" });
     });
-
+    //escape all passwords for shell
+    passwords = passwords.map((p) => {
+      let pass = p;
+      if (pass.includes('"')) pass = pass.replaceAll(/"/g, '\\"');
```
gbayasgalan, line 39: Is it better to use a library to escape all potential special characters rather than manually escaping only some of them? For example: `require('shell-escape')`

NeoPlays: they're escaped by JSON.stringify. The main problem are the double and single quotes. And because JSON.stringify already does escaping i need to remove the extra `\` introduced with `'\''`. This was all more complex then i thought tbh 😅

NeoPlays: but i think there is another problem when the password includes `\`...

gbayasgalan (approved):
lgtm 🚀

PR #2184, opened 2025-02-07
FIX: item handling on edit node page

daverolo:
- Fix bug where static "randomId" var breaks installer
- Use dynamic classes for "Config Services" items
- Avoid JS errors if no mev boost item is choosen
- Avoid JS errors if service has no dependencies
- Show "Add Connection" dialog for new KAPI service

### DIFF
--- a/launcher/src/components/UI/edit-page/EditScreen.vue
+++ b/launcher/src/components/UI/edit-page/EditScreen.vue
@@ -302,8 +302,6 @@ const setArchitecture = async () => {
   manageStore.architecture = settings.stereum?.settings.arch;
 };
 
-const randomId = computed(() => generateRandomId());
-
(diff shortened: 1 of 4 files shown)


daverolo:
@MaxTheGeeek JFYI: I removed the **static** `randomId` variable in `EditScreen.vue` because this added the **same** id on each element which did break the service add/remove/modify list on the "Edit Node" page: https://github.com/stereum-dev/ethereum-node/pull/2184/files#diff-52226eeaf9d6537c636374d24bbfb8f460d6b06f849f728ae69f81b2784c616f

NeoPlays (approved):
good job, looks good to me 🚀

there are actually some services that need the connection window. but i think they also need some work in the backend

PR #2191, opened 2025-02-18
REMOVE: legacy tag from reth

NeoPlays:
i just changed the existing script that added the tag in such a way that it removes it.


daverolo (approved):
LGTM 🚀

PR #2210, opened 2025-03-03
FEAT: Replace animation with validator import GIF in modals

daGscheid:
Add a new animation to the validator import


MaxTheGeeek (approved):
Looks good

PR #2242, opened 2025-04-06
fix: incorrect usage of `vue-loader` and `css-loader` in Webpack config

0xwitty:
I corrected an issue where the `vue-loader` and `css-loader` were being applied incorrectly in the Webpack config.
Instead of passing them as two parameters, I’ve updated it to use an array, which is the proper way to define multiple loaders in Webpack. This should fix the issue and ensure that the loaders are applied properly.


MaxTheGeeek (dismissed):
Looks great

MaxTheGeeek:
Thank you for fixing that!

daGscheid (approved):

PR #2247, opened 2025-04-08
FIX: 2FA remove issue on ubuntu-24.04

daverolo:
Fixes an issue when removing 2FA on Ubtuntu 24.04

Same as #2246 but for the "2fa-remove" task.


gbayasgalan (approved):
nice 🚀

PR #2254, opened 2025-04-29
FIX: run restart tasks in parallel, some smaller bugs

NeoPlays:
fixed module id for lido CSM
changed default chain id for testnet to the hoodi one
fixed charon oneclick setup introduced with L2 OP introducing structural changes to the OneClick-Process


NeoPlays:
haven't tested the parallel restart task yet. the automated molecule tests should give us insights on if this works

gbayasgalan (approved):
lgtm

PR #2259, opened 2025-05-20
ADD: Update Scripts for Oracle Allow List

NeoPlays:
also removes the network status (widget with last 3 epochs in slots) as the backend powering it does not work anymore and the clients report tons of warnings / errors when entering control page


gbayasgalan (approved):
lgtm

PR #2277, opened 2025-08-29
chore(ci): upgrade checkout to v5

zkpepe:
Bumps checkout to v5 for future-proofing against Node 24 runner updates.
Requires runner v2.327.1+. Workflows compile the same.

Ref: https://github.com/actions/checkout/releases/tag/v5.0.0


NeoPlays (approved):
lgtm

PR #2279, opened 2025-09-26
CHORE: update relay list

NeoPlays:
Adds `Ultra Sound Money - Filtering` relay and removes all relays that are no longer running (holesky)


daverolo (approved):
LGTM 🚀

PR #2290, opened 2026-01-14
FIX: voluntary exit messages will al be generated with the same exit epoch

NeoPlays:
Charon requires the partial exit messages to all have the same exit epoch


daverolo (approved):
LGTM 🚀

PR #2293, opened 2026-02-17
ADD: Teku Update Script and smaller improvments

NeoPlays:
bumping:
```
python to 3.12 (needed for ansible 13)
ansible to 13.3.0
molecule to 25.12.0
add molecule-plugins with 25.8.12 (replaces molecule-docker which is deprecated)
add molecule-plugins[docker]
```

added Teku Update Script which:
```
removes tags: 
--Xp2p-reworked-sidecar-recovery-enabled
--Xdata-storage-create-db-version=6

and resyncs with new db if old db is still in use
```

genesis files for prysm are now only loaded if they do not exist. (previously it was loaded on every start and stop, but github restricted the download speed to 500 kbits/s which takes for ever)


gbayasgalan (approved):
lgtm 🚀

PR #2295, opened 2026-02-18
ADD: Geth Update Script

NeoPlays:
Added update script for geth so versions between 1.16.0 and 1.17.0 remove the nodekey on update (update-changes is triggered on Stereum (Ansible) Update and when hitting update all which triggers Stereum Update)


daverolo (approved):
LGTM 🚀

PR #2308, opened 2026-08-04
CHORE: Migrate from Vue CLI to electron-vite, plus build fallout fixes

NeoPlays:
## Summary

Replaces the Vue CLI build pipeline with electron-vite, and includes four
follow-up commits that fix regressions surfaced by the migration.

The bulk of the change is the tooling swap itself (PR #2305). The remaining
commits are small, each addressing one thing that broke or was left
inconsistent after the switch.

## Changes

**1. Vue CLI to electron-vite migration** (`c3cdee97`, PR #2305)

- New electron-vite build configuration replacing the Vue CLI setup
- CI updated for the new build commands and output layout
- ESM and asset-handling fixups across a wide set of files
- Removed the obsolete 2.4.10 Ansible tasks
- Added `CLAUDE.md`

**2. Remove electron-builder from dependencies** (`bf9f93cc`)

electron-builder was listed as a runtime dependency, which broke
`npm run dist`. Moved out so packaging completes.

**3. Move web3 to dependencies** (`ecd7aca5`)

web3 was in devDependencies, so packaged builds failed at runtime with
`Cannot find module 'web3'`. It is a runtime requirement, so it belongs in
dependencies.

**4. Lock the window to a fixed, non-resizable size** (`36a809e7`)

Sets explicit window dimensions and disables resizing.

**5. Use `app.isPackaged` for dev detection** (`ebf75b1c`)

The previous dev check did not evaluate correctly in packaged builds, so the
window lock from the previous commit had no effect outside of development.
Switching to `app.isPackaged` makes the packaged and dev paths behave as
intended.

## Notes for reviewers

Commits 2 through 5 are migration fallout rather than new behavior, so they
are worth reading in order: 4 and 5 in particular only make sense together,
since 4 is inert without 5.

## Testing

- [x] `npm run dev`
- [x] `npm run dist` completes without errors
- [x] Packaged app launches and does not throw the web3 module error
- [x] Packaged window opens at the fixed size and cannot be resized
- [x] CI passes on the new build config


daverolo (approved):
lgtm 🔥

